PR #4, opened 2015-02-26
Handle the situation where you may not be able to decrease process niceness e.g. on an lxc container.

murraybd:
stgraber indicated all his lxc-containers are crashing when running unattended-upgrades because "you can't set negative nice values in an unprivileged LXC container even if you're root inside it", this fixes that by only changing the niceness if we'll be able to restore it.

### DIFF
--- a/unattended-upgrade
+++ b/unattended-upgrade
@@ -1131,7 +1131,7 @@ def main(options, rootdir=""):
 
     # FIXME: make this into a ContextManager
     # stop being nice
-    os.nice(old_priority - os.nice(19))
+    os.nice(old_priority - os.nice(0))
 
     # download what looks good
     if options.debug:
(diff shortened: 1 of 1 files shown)


a-detiste:
just my .02c

maybe this whole thing is overly complicated and not needed anymore: with Sysvinit , runing `/bin/run-parts /etc/cron.daily` with a nice value would have lead to unexpected results, like daemon restarted by `logrotate` having now a nice value; this is why each cron job must fiddle itself with the nice value.

Now, daemon restarted by systemd allways get a clean environment and exepected nice value;
a stated in their .service file.

I've for example added those lines to `cron-daily.service` on a 10 years old 1 core PC mostly used for youtube; U-U well takes ages, but user never notice nor care:

```ini
Nice=19
IOSchedulingPriority=7
CPUSchedulingPolicy=idle
```

mvo5:
Thanks, this looks great.

PR #5, opened 2015-03-26
In the apport package hook gather the right log file name.

murraybd:
I've no idea how this was overlooked, but let's fix it.


mvo5:
Thanks!

PR #7, opened 2015-05-21
Minimal patch to allow package removal on upgrade

chuhn:
minimal approach to allow package removal on upgrade, controlled via config option. 
No safety checks.
This feature is required for managing our clusters.

### DIFF
--- a/unattended-upgrade
+++ b/unattended-upgrade
@@ -524,7 +524,10 @@ def check_changes_for_sanity(cache, allowed_origins, blacklist, whitelist):
     for pkg in cache:
         if pkg.marked_delete:
             logging.debug("pkg '%s' now marked delete" % pkg.name)
-            return False
+            if not apt_pkg.config.find_b(
+                    "Unattended-Upgrade::AllowRemoveOnUpgrade",
+                    False):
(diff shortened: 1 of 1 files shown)


mvo5:
Hey Christopher! (nice github picture btw). The patch looks fine I wonder what the use-case for this is though? Could you elaborate on this a little bit? I'm a bit hesitant as this is potentially dangerous if no human checks what gets removed (might be important stuff :)

chuhn:
One example: We had Wheezy nodes upgraded to Jessie while it was still _testing_.
Then back in March gcc-4.9-base started to break gcc-4.7-base (cf. https://bugs.debian.org/770025) which was properly cleaned up `apt-get dist-upgrade` but kept back by unattended-upgrades.

We also use empty "meta" packages with lots of dependencies and the normal nightly upgrade mechanism for software deployment. This also introduces similar conflicts from time to time that should be resolved automatically.

I'm rather clueless about sensible sanity checks. gcc-4.9-base as well as gcc-4.7-base both have priority _required_, so it would be pointless to enforce something like 'Never remove important packages'.

PR #10, opened 2015-06-24
display translations in proper locale

anarcat:
without this, translations would get encoded using the ascii codec
which would break with an infamous UnicodeEncodeError exception. by
using lgettext(), the proper encoding is automatically chosen on
output which seems to be a good practice here.

notice that i had to change from the string formatting operator to
arguments to the logging function in a specific case, otherwise i
would get the following exception:

TypeError: unsupported operand type(s) for %: 'bytes' and 'list'

This is a weird Python 3 specific exception, but it makes sense: the
string is encoded to bytes already and the formatting operator doesn't
want to take any chances. the proper way to fix this is to use the
builtin formatting of the logging function. i have done this in the
one place where i still had, but it's possible that i missed other possible places.

### DIFF
--- a/unattended-upgrade
+++ b/unattended-upgrade
@@ -41,7 +41,7 @@ import sys
 
 
 from email.message import Message
-from gettext import gettext as _
+from gettext import lgettext as _
 from io import StringIO
 from optparse import (
     OptionParser,
@@ -1052,7 +1052,7 @@ def main(options, rootdir=""):
     logging.info(_("Starting unattended upgrades script"))
 
     # display available origin
-    logging.info(_("Allowed origins are: %s") % allowed_origins)
+    logging.info(_("Allowed origins are: %s"), allowed_origins)
 
     # check if the journal is dirty and if so, take emergceny action
     # the alternative is to leave the system potentially unsecure until
--- a/unattended-upgrade
+++ b/unattended-upgrade
@@ -398,8 +398,8 @@ def upgrade_normal(cache, pkgs_to_upgrade, logfile_dpkg, verbose):
         logging.info(_("All upgrades installed"))
     else:
         logging.error(_("Installing the upgrades failed!"))
-        logging.error(_("error message: '%s'") % error)
-        logging.error(_("dpkg returned a error! See '%s' for details") %
+        logging.error(_("error message: '%s'"), error)
+        logging.error(_("dpkg returned a error! See '%s' for details"),
                       logfile_dpkg)
     return res
 
@@ -477,8 +477,8 @@ def upgrade_in_minimal_steps(cache, pkgs_to_upgrade, blacklist,
             if verbose:
                 logging.exception("Exception happened during upgrade.")
             logging.error(_("Installing the upgrades failed!"))
-            logging.error(_("error message: '%s'") % e)
-            logging.error(_("dpkg returned a error! See '%s' for details") %
+            logging.error(_("error message: '%s'"), e)
+            logging.error(_("dpkg returned a error! See '%s' for details"),
                           logfile_dpkg)
             return False
         to_upgrade = to_upgrade - set(smallest_partition)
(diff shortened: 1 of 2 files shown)


anarcat:
forgot to mention this is a release-critical bug in debian jessie: http://bugs.debian.org/cgi-bin/bugreport.cgi?bug=789277

mvo5:
Thanks a bunch, I'm looking into this and its surprisingly tricky to do, especially since I want to keep py2 and py3 compatibility. I created a bugfix/bts789277 branch that looks good for py3. Feedback welcome for this one.

a-detiste:
As a workaround you could add a `LANG=fr_CA.UTF-8` at the top of `/etc/crontab` ?

mvo5:
Thanks, after looking at this again it all makes sense, thanks a lot for this branch!

anarcat:
note that i am now getting this with the patch:

```
verbose level 1
power status (255) undetermined, continuing
system is on main power.
power status (255) undetermined, continuing
system is on main power.
check_stamp: interval=86400, now=1437537600, stamp=1437451200, delta=86400 (sec)
download updated metadata (success).
send dbus signal (success)
check_stamp: interval=0
download upgradable (not run)
check_stamp: interval=86400, now=1437537600, stamp=1435464000, delta=2073600 (sec)
Traceback (most recent call last):
  File "/usr/bin/unattended-upgrade", line 1321, in <module>
    main(options)
  File "/usr/bin/unattended-upgrade", line 1173, in main
    pkgname_from_deb(item.destfile))
TypeError: unsupported operand type(s) for %: 'bytes' and 'str'
unattended-upgrade (error)
check_stamp: interval=0
autoclean (not run)
skip aging since MaxAge is 0
```

not sure if it's a regression...

a-detiste:
me too, not getting mails anymore

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=793663


```
2015-08-01 07:28:40,602 INFO Initial whitelisted packages: 
2015-08-01 07:28:40,602 INFO b'D\xc3\xa9marrage du script de mise \xc3\xa0 niveau automatique'
2015-08-01 07:28:40,603 INFO b'Les origines permises sont\xc2\xa0: ['origin=Debian,codename=stretch,label=Debian-Security', 'o=Debian,a=testing', 'o=a_detiste', 'l=Google,a=stable', 'o=Debian,a=unstable', 'o=Debian,a=experimental']'
2015-08-01 07:30:53,210 INFO Packages that will be upgraded: bash bind9-host debbindiff dnsutils ffmpegthumbs ghostscript hugin hugin-data hugin-tools libbind9-90 libdns100 libfontenc1 libgs9 libgs9-common libisc95 libisccc90 libisccfg90 liblwres90 libosgearth3 libosgearthannotation3 libosgearthfeatures3 libosgearthqt3 libosgearthsymbology3 libosgearthutil3 libpoppler-glib8 libpoppler-qt4-4 libpoppler-qt5-1 libpoppler46 osgearth-data poppler-utils qprint twolame
2015-08-01 07:30:53,210 INFO b'\xc3\x89criture du journal de dpkg dans \xc2\xab\xc2\xa0/var/log/unattended-upgrades/unattended-upgrades-dpkg.log\xc2\xa0\xc2\xbb'
```

anarcat:
looks like latest jessie fixed my regression here...

mvo5:
On Sat, Aug 01, 2015 at 01:00:03PM -0700, Alexandre Detiste wrote:
> me too, not getting mails anymore
> 
> https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=793663

I replied in the debian bug as well. I could not reproduce this just
yet, if you can, I wonder what:
"""
print(type(res))
print(type(_(""))
"""
returns right before the line that crashes. The error looks like _()
returns a byte instance instead of a string. However looking at the
gettext.py implementation from python3.4 it looks like its always
encoding it to a string. Help how I can reproduce this would be
great. 

Thanks,
 Michael

a-detiste:
OK, but I'm still stuck with the old apt from experimental for a while because of GCC transition (could this come from libapt ?)

PR #12, opened 2015-07-02
update french translation

a-detiste:
@anarcat : can you review my translation ? currently having bits of English & French in the log feels weird.


mvo5:
Fwiw, I'm happy to merge this but I can not review it, my french is not good enough for that :)

a-detiste:
I sent a mail to debian-l10n-french@lists.debian.org & to original translator, but got no reply. You could as well only cherry pick the first commit 7630e8c to help translations in other languanges, but it's not like there is a hurry.

anarcat:
looks good except for one thing, and that's a maybe.

PR #14, opened 2015-09-07
Add [package on hold] to mail subject

Jerome-github:
This PR adds a [package on hold] "flag" in the mail subject to catch attention, just like already existing [reboot required].

The rationale is that an admin may want to get all emails but only read those that need manual action.

See my comment [here](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=625847#35).

Feel free to ask for a rework if you like the idea but not the implementation.

Note that this change removes string

    msgid "[reboot required] unattended-upgrades result for '%s'"

and adds

    msgid "[reboot required]"
    msgid "[package on hold]"

the first can be easily stripped from existing translation file, so no real regression here.

### DIFF
--- a/unattended-upgrade
+++ b/unattended-upgrade
@@ -806,13 +806,14 @@ def send_summary_mail(pkgs, res, pkgs_kept_back, mem_log, dpkg_log_content):
         return
     # Check if reboot-required flag is present
     logging.debug("Sending mail to '%s'" % to_email)
+    subject = _("unattended-upgrades result for '%s': %s") % (host(), res)
+    flags = ""
     if os.path.isfile(REBOOT_REQUIRED_FILE):
-        subject = _(
-            "[reboot required] unattended-upgrades result for '%s': %s") % (
-            host(), res)
-    else:
-        subject = _("unattended-upgrades result for '%s': '%s'") % (
-            host(), res)
+        flags += _("[reboot required]")
+    if pkgs_kept_back:
+        flags += _("[package on hold]")
+    if flags:
+        subject = flags + " " + subject
     body = _("Unattended upgrade returned: %s\n\n") % res
     if os.path.isfile(REBOOT_REQUIRED_FILE):
         body += _(
--- a/unattended-upgrade
+++ b/unattended-upgrade
@@ -805,15 +805,16 @@ def send_summary_mail(pkgs, res, pkgs_kept_back, mem_log, dpkg_log_content):
             "Unattended-Upgrade::MailOnlyOnError", False)):
         return
     # Check if reboot-required flag is present
+    reboot_flag_str = _(
+        "[reboot required]") if os.path.isfile(REBOOT_REQUIRED_FILE) else ""
+    # Check if packages are kept on hold
+    hold_flag_str = _("[package on hold]") if pkgs_kept_back else ""
     logging.debug("Sending mail to '%s'" % to_email)
-    subject = _("unattended-upgrades result for '%s': %s") % (host(), res)
-    flags = ""
-    if os.path.isfile(REBOOT_REQUIRED_FILE):
-        flags += _("[reboot required]")
-    if pkgs_kept_back:
-        flags += _("[package on hold]")
-    if flags:
-        subject = flags + " " + subject
+    subject = _(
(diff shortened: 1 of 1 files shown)


mvo5:
Thanks! I like the spirit of the change a lot.

I'm a bit hesitant about the implementation. Generally for i18n strings its recommended to not concat them because some languages may need to change the order of the strings. I.e. in some languages the "[reboot-required" part may need to go to the end instead of the beginning. I have no idea if that is actually a issue for flags like [reboot-required] but I try to avoid it in general. So if its not too much work it would be nice if you could look if it can be done as separate strings. 

Thanks!
 Michael

Jerome-github:
Hi. Glad you like it.

I understand the general recommendation about string concatenation. As you say, this is a special case as it is flags, not parts of a sentence, so I didn't think it would be an issue. But now that you raise the point, I'm thinking it might be wrong for people writing from right to left, for instance.

Creating one string for each flag combination would be feasible for 2 flags, that's only 4 combinations, but it would be anything but elegant and it would become a nightmare if more flags were to be added.

We could do something along the lines of this (untested):

    flags = ""
    if os.path.isfile(REBOOT_REQUIRED_FILE):
        flags += _("[reboot required]")
    if pkgs_kept_back:
        flags += _("[package on hold]")
    if flags:
        subject = _("{2} unattended-upgrades result for '{0}': {1}" '%s': %s").format( % (host(), res, flags)
    else:    
        subject = _("unattended-upgrades result for '{0}': {1}" '%s': %s").format( % (host(), res)

We could use a common statement and a common string, but then we would have an issue with the space between the flags and the "unattended-upgrades..." text. It looks nicer with a space between the flags and the text, but it requires duplicating the string (unless there is another way?).

Do you like this implementation? Would you like me to submit it?

(Note to self: I shall also fix/adapt/extend the comment that reads "Check if reboot-required flag is present".)

Jerome-github:
The proposal in my former comment allows the translator to choose whether flags should be left or right from the subject, but does not affect the order of the flags themselves.

Assuming the order would be important (like most important first, first being left or right depending on the language), we may want to enforce this as well.

It could be done this way (also untested):

subject = _("unattended-upgrades result for '{0}': {1}" '%s': %s").format(host(), res)

    flags = []
    if pkgs_kept_back:
        flags.append(_("[package on hold]"))
    if os.path.isfile(REBOOT_REQUIRED_FILE):
        flags.append(_("[reboot required]"))
    for flag in flags:
        subject = _("{1} {0}").format(subject, flags)

This would yield 

    "[reboot required] [package on hold] unattended-..."

and the translator could translate `"{1} {0}"` into `"{0} {1}"` to get 

    "unattended-... [päckagê on hõld] [rëboõt reqûïrëd]"

Pros: 

- It might stick better to recommendations (less concatenation)
- Only one instance of "unattended-..." string translation to maintain

Cons:

- The `"{1} {0}"` trick looks a bit cryptic, especially for the translator. Maybe we can get away with it by just adding a helpful comment in the code...

I don't have much experience in internationalization. There could be better ways.

mvo5:
Oh, I like the "{0}" idea! Could we do sometthing like:

   "{hold_flag}{reboot_flag} unattended-upgrades result for '{machine}': {result}". format(hold_flag=hold_flag_str, reboot_flag=reboot_flag_str, machine=machine_str, result=result)

? This way its easy for the translators to know what the {} means, easy to reorder and if hold_flag_str is empty the string looks just the same. Does that sounds like a sensible idea?

Jerome-github:
Excellent! Named parameters is much better.

The only thing that bugs me is the leading space if both hold_flag and reboot_flag are empty.

Maybe we could [`strip()`](https://docs.python.org/3.4/library/stdtypes.html#str.strip) the string after translation.

    _("{hold_flag}{reboot_flag} unattended-upgrades result for '{machine}': {result}").format(hold_flag=hold_flag_str, reboot_flag=reboot_flag_str, machine=machine_str, result=result).strip()

I don't think this breaks the "no messing around with strings" principle.

I'll do that and update this PR.

Jerome-github:
Hi Michael.

I updated the branch as per your suggestions. Much better now. I should have figured that out straight away.

Again, please tell me if a rework is needed.

mvo5:
Thanks!

PR #15, opened 2015-09-08
Add debconf question to set an origin pattern

jvalleroy:
Please consider merging this change, which adds a debconf question to set the default (uncommented) line of Origins-Pattern.

Here are some details about the change:
1. 50unattended-upgrades will be installed under /usr/share/unattended-upgrades, instead of under /etc.
2. A temporary copy of this file is made, and the default Origins-Pattern line is replaced with the debconf answer (or default value).
3. ucf is used to merge the modified 50unattended-upgrades with the existing file under /etc.

This change is needed by the FreedomBox (a Debian pure blend), in order to offer smooth upgrades to our users. We would like to use unattended-upgrades to handle all upgrades (including upgrading unattended-upgrades itself, of course). So we need to do all of our needed configurations through debconf, so we can avoid conffile prompts while still receiving any updates to the packaged config file.


a-detiste:
why not simply provide `51freedombox` to override previous file ?

jvalleroy:
@a-detiste Because if we provide an override file, we would have to maintain it, forever. It's better if we can have unattended-upgrades maintain its own configuration, while still providing this flexibility.

I also just realized, that the debconf question I proposed here may be overly broad. I believe it could be reduced to a simple boolean "Install only security upgrades by default?", and that would still meet our needs. How does this sound?

mvo5:
Thanks for the branch. I love the idea of moving it to ucf and avoiding the conffile prompts.

PR #16, opened 2015-09-23
Speed up sanity check by filtering cache.

HyperBaton:
The apt.cache.Cache.get_changes() method returns only packages that are not marked as keep. The sanity check does not care for packages that are marked as keep, so it seems to make sense to filter the cache first to avoid unneeded loops.

On a test system with 140 upgradeable packages, this reduced the runtime of unattended-upgrade from 4m32s to 1m5s (without installation of packages).


mvo5:
Yay, this looks great! Thanks a lot :)

HyperBaton:
Cheers, thanks Michael! I'm not very familiar with Debian policy regarding which patches get applied in a next jessie release. Would this be something that could get merged in jessie 8.3?

PR #19, opened 2015-12-21
Add Unattended-Upgrade::Remove-New-Unused-Dependencies

mvo5:
The new Unattended-Upgrade::Remove-New-Unused-Dependencies will
remove anything that is no longer required after an
unattended-upgrade run. It defaults to "true" to ensure that
e.g. old kernels are removed and /boot will not fill up.

Closes: LP: #1357093


GhostLyrics:
I assume you're doing this to avoid uninstalling leftover dependencies from manual apt runs. Do you intend to keep both remove-unused options in the long term?

Also I am sure you are aware that. you will still fill up /boot if the user upgraded using previous versions of the package and has lots of old kept kernels still lying around.

mvo5:
@GhostLyrics Indeed, the rational is that at least from now on the /boot problem is gone. And yes, I know that this won't solve the problem for existing /boot overflowing partitions.

jarnos:
@mvo5 I think you made and error in changelog, when you tell the related bug is LP: #1267059. I think #1054927 is the right one.

jarnos:
I suppose for the fix to work, unattended-upgrades has to be enabled. Will that be enabled by default? In /etc/apt/apt.conf.d/50unattended-upgrades it could be configured to install no updates by default i.e. nothing in Unattended-Upgrade::Allowed-Origins so that user can install all updates manually by update-manager and still avoid /boot from becoming full.

jarnos:
As for Remove-New-Unused-Dependencies, does it remove the newest kernel that is older than the installed one? That is probably the booted one then?! Wouldn't it be better to remove (or even purge) the oldest one?

jarnos:
I question the justification of this setting. See  LP: #1624644

PR #20, opened 2015-12-22
Set debconf value for auto updates based on current configuration.

### LINKED ISSUE #18: Should populate debconf database with settings from disk, if they exist
I ran into this situation:
1. Installed unattended-upgrades, using default debconf settings (enabled_auto_updates == false).
2. Manually created /etc/apt/apt.conf.d/20auto-upgrades with Update-Package-Lists == "1" and Unattended-Upgrade == "1".
3. Ran "apt upgrade" which happened to upgrade unattended-upgrades. This led to a ucf prompt asking me whether to keep my 20auto-upgrades file, or replace it with 20auto-upgrades-disabled.

This can be avoided by having config check if 20auto-upgrades exists, and reading the values to determine the current setting of enable_auto_updates. It should set this value in debconf before running db_input.

jvalleroy:
This fixes #18.

Also, with this change, when the user runs ```dpkg-reconfigure unattended-upgrades```, the current setting will match the setting from the config file, regardless of the value left in the debconf cache.

### DIFF
--- a/debian/config
+++ b/debian/config
@@ -1,7 +1,18 @@
 #!/bin/sh -e
 
+AUTO_UPGRADE="/etc/apt/apt.conf.d/20auto-upgrades"
+
 # Source debconf library.
 . /usr/share/debconf/confmodule
+
+# Load configuration from disk, if it exists.
+if [ -e "$AUTO_UPGRADE" ]; then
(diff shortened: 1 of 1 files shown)


mvo5:
Nice, thanks a bunch!

PR #21, opened 2016-01-05
Allow to configure sender email for sendmail

Baertierchen:
This allows to specify the email address of the sender. At least when using sendmail.

I needed this as I use kuvert to sign and encrypt outgoing mails. Kuvert requieres that the from field of the email is set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,6 +191,15 @@ Supported Options Reference
  Unattended-Upgrade::Mail "user@example.com";
  ```
 
+* `Unattended-Upgrade::Sender` - string (default:"root")
+
+ Use the specified value in the "From" field of outgoing mails.
+ 
+  Example:
+ ```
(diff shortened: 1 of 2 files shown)


mvo5:
Thanks a lot for this branch! This looks fine. It would be nice if you could document the option in the README.md.

mvo5:
Oh, and mailx seems to also support a "from" override, so it would be nice if send_mail_using_mailx() could also be updated, the option is "-r" in this case.

Thanks!
 Michael

Baertierchen:
Just added the documentation and the implementation for mailx. The mailx implementation, however, is not tested as I have no mailx installation at hand. Also, I don't know why the tests failed ;)

mvo5:
Thanks! I merged this now and added a small fix for the failing test in 0527c28bd20c20763f7e7a929196c205ff0c095d

PR #23, opened 2016-01-27
bugfix: non-ascii chars in dpkg log file crashes unattended upgrades

cletomartin:
This is a minor tweak that I think fix this bug: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=812857

### DIFF
--- a/unattended-upgrade
+++ b/unattended-upgrade
@@ -29,6 +29,7 @@ import fcntl
 import fnmatch
 import gettext
 import grp
+import io
 import locale
 import logging
 import re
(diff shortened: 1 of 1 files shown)


lafrech:
Travis check failed because of `encoding`, which is Python 3 only. More info about this here: http://stackoverflow.com/questions/10971033/backporting-python-3-openencoding-utf-8-to-python-2.

Travis tests for both Python 2 and 3.

    # test with both py2 and py3 for now and also use coverage for the py2

I'm wondering about the need to support Python2. Is there a platform targetted by `unattended-upgrades` that still does not have Python 3?

GhostLyrics:
As long as potential backports to `precise` might be necessary, probably yes. Afair the UU package depends on Python 2 there. Not entirely sure about `trusty`.

cletomartin:
Thanks for the feedback! I guess I can use `io.open` instead then, right? https://docs.python.org/2/library/io.html#io.open

mvo5:
Hey @cletomartin - nice to see you. Thanks for the branch.

PR #26, opened 2016-02-12
Update changelog

jarnos:
The bug is #1054927 instead of  #1267059

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -10,7 +10,7 @@ unattended-upgrades (0.89) unstable; urgency=medium
   * Add `Unattended-Upgrade::Remove-New-Unused-Dependencies` that
     defaults to "yes". This ensures that older kernel
     get automatically cleaned up and /boot will not overflow.
-    LP: #1267059
+    LP: #1054927
   * Remove downloaded deb packages after successful installs.
     This can be controlled via the option:
     `Unattended-Upgrade::Keep-Debs-After-Install`


mvo5:
Its both bugs if I'm not mistaken? So maybe adding the older bugnumber instead of replacing it?

jarnos:
Launchpad says Bug #1267059 has been fixed already in unattended-upgrades (0.76ubuntu1.2). That bug is not about what setting is default.

Bug #1294195 tells Unattended-Upgrade::Remove-Unused-Dependencies is now default, but
you tell in changelog that `Unattended-Upgrade::Remove-New-Unused-Dependencies` defaults to "yes" now. Isn't that different thing?

Bug #1054927 is about making making the removal of unneeded kernels automatic (at least if unattended-upgrades is used to install kernels automatically).

Is unattended upgrades configured to work by default in upcoming Ubuntu? I mean if user has not set automatic installation of security updates on in software-properties-gtk, do unneeded kernels get automatically removed even then?  A solution could be https://bugs.launchpad.net/ubuntu/+source/update-manager/+bug/1389620/comments/4
There should be a GUI way to configure what is in Unattended-Upgrade::Allowed-Origins block then, I think. Another solution is the one requested in Bug #1389620.

Maybe package update-manager does not need to be fixed for Bug #1054927, but for Bug #1389620 and/or for arguably more general Bug #1357093, if at all.

mvo5:
Indeed, thanks a bunch! I forgot about the subtle differences.

mvo5:
As for automatic updates in Ubuntu 16.04 - yes, the plan is to have that enabled by default and also to enable automatic removals of unused kernels. So that part should be good. The update-manager where people apply updates manually still needs work though.

jarnos:
Kernels installed by update-manager are marked as automatically installed, except in Trusty (Bug #1492709) so couldn't unattended upgrades handle removing unneeded packages automatically even when security updates are installed by update-manager? /boot may be filled by e.g. manually installed upstream kernels, but they are installed by other means than by update-manager.

jarnos:
I just tested that it is possible to remove unneeded kernels by unattended upgrades even if kernels are installed in update-manager in Wily.

PR #27, opened 2016-02-25
Feature/patch day

mvo5:
Add configuration to allow to specify at which days of the week updates should be applied. This makes it easier to spread out updates in larger installations to ensure that a single faulty update does not kill all machines at once.


mbearup:
This change looks good to me, no issues.

PR #31, opened 2016-09-18
Speed up sanity check by filtering cache.

HyperBaton:
This pull request was already done on the debian/jessie branch but I would like to get this in the master branch as well.

The apt.cache.Cache.get_changes() method returns only packages that are not marked as keep. The sanity check does not care for packages that are marked as keep, so it seems to make sense to filter the cache first to avoid unneeded loops.

On a test system with 140 upgradeable packages, this reduced the runtime of unattended-upgrade from 4m32s to 1m5s (without installation of packages).

### DIFF
--- a/unattended-upgrade
+++ b/unattended-upgrade
@@ -553,7 +553,7 @@ def is_pkgname_in_whitelist(pkgname, whitelist):
 def check_changes_for_sanity(cache, allowed_origins, blacklist, whitelist):
     if cache._depcache.broken_count != 0:
         return False
-    for pkg in cache:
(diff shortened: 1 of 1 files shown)


HyperBaton:
I'm not too familiar with the automated tests. I suppose the get_changes() method should be added to the MockCache?

HyperBaton:
Closing due to inactivity.

a-detiste:
@mvo5 please apply this one to master

On some old laptop I've left in a drawer, when powered up UU can take up to one hour
computing the update before getting to the `dpkg` stage.

PR #35, opened 2016-11-07
Update 50unattended-upgrades.Ubuntu

murraybd:
Modify data/50unattended-upgrades.Ubuntu such that the release pocket is an allowed origin so that security updates with a new dependency will be upgraded and the new dependency will be installed.  This fixes Launchpad bug 1624641 and is being used in Ubuntu 16.04 and 16.10.


mvo5:
Thank you!

PR #38, opened 2016-12-08
ensure network and local-fs are available

murraybd:
For packages to be updated on shutdown unattended-upgrades needs access to the network and local-fs see https://bugs.launchpad.net/ubuntu/+source/unattended-upgrades/+bug/1618900


mvo5:
Thanks!

PR #44, opened 2017-01-04
Do not mark packages for deletion if using --dry-run

murraybd:
This fixes a bug reported in Launchpad.  I tried really, really hard to write a test case for this but from what I could tell TestRemoveUnused is not idempotent and running my new test by itself would pass but some others would fail.  I set it up to run the failing ones by themselves and they'd pass.  I think putting the new test in a separate class worked but that seemed terrible.  Anyway, I couldn't figure it out and have been sitting on this for a couple of months.  Now, I figure its better to just get the bug fixed.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,10 @@
+unattended-upgrades (0.93.2) UNRELEASED; urgency=medium
+
+  * Do not mark packages for deletion / autoremoval if unattended-upgrades is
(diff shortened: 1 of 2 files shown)


murraybd:

rbalint:
This change got merged in https://github.com/mvo5/unattended-upgrades/pull/59 without the changelog update. Thanks!

PR #45, opened 2017-01-06
improve the changes sanity check to cache more insane cases

murraybd:
This is pretty well detailed in https://bugs.launchpad.net/ubuntu/+source/unattended-upgrades/+bug/1654070 but for some reason the new version of libpulse0 is out of date in my cache and subsequently pulseaudio-module-gconf (which depends on libpulse0) won't update.  However, I kept getting emails that the package was being upgraded, even though it wasn't, and decided to find why.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,11 @@
+unattended-upgrades (0.93.2) UNRELEASED; urgency=medium
+
+  * When performing a sanity check for packages to install or upgrade return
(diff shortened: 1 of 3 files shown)


murraybd:

rbalint:
This change got merged in https://github.com/mvo5/unattended-upgrades/pull/59 without the changelog update. Thanks!

Issue #46, opened 2017-01-27
--delay argument doesn't work in unattended-upgrade-shutdown


karibou:
OptionParser returns --delay value as a string.

if (time.time() - start_time) > options.delay * 60:

This fails when --delay is used since options.delay is a string.

PR #47, opened 2017-01-27
Fix --delay type in unattended-upgrade-shutdown

karibou:
delay value is returned as a string which makes the calculation
fail (Closes: #46)

Signed-off-by: Louis Bouchard <louis.bouchard@canonical.com>


murraybd:

PR #54, opened 2017-05-24
Do not automatically upgrade the development release of Ubuntu

murraybd:
With unattended-upgrades being enabled by default in Ubuntu a side effect is the installing of any new packages when someone is running the development release.  This fixes that by adding a DevRelease configuration option which defaults to false.

The problem is documented here:

https://bugs.launchpad.net/ubuntu/+source/unattended-upgrades/+bug/1649709

### DIFF
--- a/data/50unattended-upgrades.Ubuntu
+++ b/data/50unattended-upgrades.Ubuntu
@@ -19,6 +19,10 @@ Unattended-Upgrade::Package-Blacklist {
 //	"libc6-i686";
 };
 
+// This option will controls whether the development release of Ubuntu will be
+// upgraded automatically.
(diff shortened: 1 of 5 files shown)


rbalint:
I think if one does not want to run unattended-upgrade on the development release she/he should just remove it. I think the reported issue is not a bug and u-u does the right thing by upgrading all packages and there is no other way of getting security fixes during the development period.

rbalint:
Superseded by https://github.com/mvo5/unattended-upgrades/pull/54 which included an updated version of this change.

rbalint:
@murraybd Thanks for the fix!

PR #56, opened 2017-06-06
allow configuration of a pre-execution hook for arbitrary conditions

d10r:
On my notebook, I wanted automatic updates to be done only when connected to Ethernet and AC Power.
Since I didn't find a way to do this with unattended-upgrades as is, I added a config option which allows to make execution dependent on arbitrary conditions by executing a given script and checking its return value.

Usage described in `README.md`

As an example, that's the script I'm using it with (tailored to my system):
```
#!/bin/bash

set -e
set -u

ETH_CONNECTED=false
AC_CONNECTED=false

# check Ethernet connected
if [[ `cat /sys/class/net/eth0/operstate` == "up" ]]; then
  ETH_CONNECTED=true
fi

# check AC connected
if [[ `cat /sys/devices/platform/smapi/ac_connected` == 1 ]]; then
  AC_CONNECTED=true
fi

if $ETH_CONNECTED && $AC_CONNECTED; then
  exit 0
fi

exit 1
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,6 +261,15 @@ Supported Options Reference
  ```
  The default is an empty list which means updates are applied every day.
 
+* `Unattended-Upgrade::Special-Conditions-Script` - path to a script on which to make execution dependent
+
+ Allows the definition of arbitrary conditions for update execution.
+ If set, the given script is executed. If it returns 0, unattended-upgrades proceeds, otherwise it stops. 
+
+ Example - apply updates only with Ethernet and AC Power connected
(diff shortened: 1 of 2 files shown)


rbalint:
In typical configurations the processes calling unattended-upgrade implement that thus i think there would be little value in adding the feature to u-u, too. 
https://anonscm.debian.org/git/apt/apt.git/commit/?id=51d659e7d8cdce59f910eceeee68e2c2afdb70d4

PR #62, opened 2017-07-15
Lock handling fixes and protecting u-u from being killed during shutdown

rbalint:


rbalint:
The changelog does not contain all Debian BTS bug numbers which would be closed by the upload.

debian/postinst
```diff
@@ -34,10 +34,18 @@ case "$1" in
...
 	    ucfr unattended-upgrades "$CONFIG"
+	    NEWDROPIN="/usr/share/unattended-upgrades/unattended-upgrades-shutdown.conf"
+	    mkdir -p $(dirname "$DROPIN")
+	    ucf --three-way --debconf-ok "$NEWDROPIN" "$DROPIN"
+	    ucfr unattended-upgrades "$DROPIN"
+	    if [ -d /run/systemd/system ]; then
+	        systemctl --system daemon-reload >/dev/null || true
+	    fi
         elif [ "${RET}" = "false" ] && [ -e "$CONFIG" ]; then
             # disable again
```
rbalint: disabling the drop-in is missing here, I'll add it

Issue #52, opened 2017-05-15
Blacklisted packages not upgraded but not in pkgs_kept_back


lafrech:
Hi.

I set a few package in blacklist:

```python
Unattended-Upgrade::Package-Blacklist {
        "vim";
        "libc6";
        "libc6-dev";
        "libc6-i686";
        "dovecot";
        "roundcube";
        "postgresql";
        "mysql-server";
};
```

Some of these currently have upgrades available.

They are hold back as expected.

However, I expected them to appear in the "Packages with upgradable origin but kept back" part of the mail, and they don't.

The only package that appears in the mail is `libc-dev-bin`.

I can't tell whether this is an issue or I'm just misunderstanding something.

I thought the point of this section of the mail was to inform the user that some packages could be upgraded. This does not seem to be the case. If I don't launch `apt-get`/`aptitude` myself, I won't be notified.

I am indeed notified about the libc update (the mail mentions `libc-dev-bin`) but not about postgres and vim updates, for instance.

(I was using Debian's stable version (0.83.3.2+deb8u1). I upgraded to latest Debian package (0.93.1+nmu1) and the behavior is the same.)

GhostLyrics:
The section in the mail is to be read as "apt tries to upgrade the package but it was a failure.( Or it could not be upgraded due to dependency issues)". 

It is not what you expect. What you expect is "apt was not allowed to upgrade this package."

lafrech:
Thanks @GhostLyrics, it makes more sense to me, now.

- `libc-dev-bin` can't upgrade because it `replaces` a version of `libc6-dev` that itself can not be updated due to blacklisting, hence the "failure or dependency issue" and the message.

- The other packages can't be replaced because they are blacklisted more directly.

Wouldn't it be more consistent to notify the admin in both cases?

AFAIU, the point of unattended-upgrades is to avoid launching apt-get/aptitude regularly and let it manage upgrades, either updating automatically, either notifying the admin. This purpose gets defeated if it silently ignores blacklisted upgrades.

There are other tools to be notified but using them would be redundant and the admin would be notified for every upgradable package, even the 99% that are automatically upgraded.

Would it make sense to notify about all upgradable but not upgraded packages?

- failures
- dependency issues
- blacklisted
- origin not allowed

[...]

PR #63, opened 2017-07-19
Blacklisted packages are now added to the pkgs_kept_back list - fix for #52 ?

jayaich:
This pull request simply adds the relevant blacklisted package to the pkgs_kept_back list. To do this though it also required the calling functions to be passed the list.
Tested and seems to work okay when only blacklisted packages are available for updating, and for when there are blacklisted and 'ordinary' packages available (packages which are neither black nor white listed).
(One trivial typo also included.)

### DIFF
--- a/unattended-upgrade
+++ b/unattended-upgrade
@@ -1112,7 +1112,8 @@ def calculate_upgradable_pkgs(cache, options, allowed_origins,
                 pkg.name, getattr(pkg.candidate, "origins", [])))
 
         if (pkg.is_upgradable and
-                not is_pkgname_in_blacklist(pkg.name, blacklisted_pkgs, pkgs_kept_back) and
+                not is_pkgname_in_blacklist(pkg.name, blacklisted_pkgs,
+                                            pkgs_kept_back) and
                 is_pkgname_in_whitelist(pkg.name, whitelisted_pkgs) and
                 is_allowed_origin(pkg.candidate, allowed_origins)):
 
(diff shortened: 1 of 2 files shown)


unattended-upgrade
```diff
@@ -395,7 +395,7 @@ def match_whitelist_string(whitelist, origin):
             res &= fnmatch.fnmatch(origin.codename, value)
         else:
             raise UnknownMatcherError(
-                "Unknown whitelist entry for macher '%s' (token '%s')" % (
+                "Unknown whitelist entry for matcher '%s' (token '%s')" % (
```
mvo5: Nice catch, thank you!

mvo5:
A test for this would be great (or just updating an existing one). Please let me know if that would be something you would be interested in doing :)

jayaich:
I'm just taking a look at the tests. I see that whitelisting is checked, but perhaps not blacklisting. I'm no python guru, so I'll see if I can do anything but it may well come back your way.

jayaich:
As the main part of this fix (#52) just ensures that blacklisted packages are in the pkgs_held_back list, and this is used within the summary email message, I have modified the 'test/test_mail.py' file to include a couple of tests. Both tests ensure that the subject line is correct, and that the relevant text line is in the mail message. The second test ensures that it is still all correct even if there are no packages updated at that time.

mvo5:
@jayaich Yay, thanks a lot! \o/

PR #64, opened 2017-07-22
Protect  unattended-upgrades from stopping early

rbalint:
Also protect the phase fixing broken installs

### DIFF
--- a/data/50unattended-upgrades.Debian
+++ b/data/50unattended-upgrades.Debian
@@ -54,7 +54,7 @@ Unattended-Upgrade::Package-Blacklist {
 //Unattended-Upgrade::AutoFixInterruptedDpkg "false";
 
 // Split the upgrade into the smallest possible chunks so that
-// they can be interrupted with SIGUSR1. This makes the upgrade
+// they can be interrupted with SIGTERM. This makes the upgrade
 // a bit slower but it has the benefit that shutdown while a upgrade
 // is running is possible (with a small delay)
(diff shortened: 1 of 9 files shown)


debian/control
```diff
@@ -14,7 +14,7 @@ XS-Testsuite: autopkgtest
 Package: unattended-upgrades
 Architecture: all
 Depends: ${shlibs:Depends}, ${misc:Depends}, debconf, python3, 
- python3-apt, apt-utils, apt, ucf, lsb-release, lsb-base (>= 3.2-14),
+ python3-apt, apt-utils, apt (>= 1.4.2~), ucf, lsb-release, lsb-base (>= 3.2-14),
```
mvo5: This is slightly problematic - I would like unattended-upgrades to be trivially backportable as far as possible. This is why it runs under both py2 and py3 and does not use advanced python-apt features. If we can avoid making this a hard dependencies and instead do runtime detection of whatever it is we need, I would prefer that. I'm not far enough into the PR yet to see why this is needed :)

rbalint: I'm proposing a patch against apt instead and u-u then can drop the versioned dependency.
Please don't merge before I update this PR.

rbalint:
Please follow this Launchpad bug to see if the apt patch gets accepted, this PR relies on the apt KillMode and timeout changes:
https://bugs.launchpad.net/ubuntu/+source/unattended-upgrades/+bug/1690980

PR #67, opened 2017-07-27
There is no such Raspbian-Security label

nachoparker:
According to [this](https://www.raspberrypi.org/forums/viewtopic.php?f=66&t=82863) Raspbian thread and my own testing, there is no _Raspbian-Security_ label. 

See also [this](https://github.com/jnv/ansible-role-unattended-upgrades/pull/20/commits/43830851211a87fcada9a43e3e22e8d8c3e61656)

> In Raspbian, it is only possible to update all packages from the default repository, including non-security updates, or updating none.

### DIFF
--- a/data/50unattended-upgrades.Raspbian
+++ b/data/50unattended-upgrades.Raspbian
@@ -8,7 +8,7 @@
 // file, but several aliases are accepted.  The accepted keywords are:
 //   a,archive,suite (eg, "stable")
 //   c,component     (eg, "main", "contrib", "non-free")
-//   l,label         (eg, "Rapsbian", "Raspbian-Security")
+//   l,label         (eg, "Rapsbian", "Raspbian")
 //   o,origin        (eg, "Raspbian", "Unofficial Multimedia Packages")
 //   n,codename      (eg, "jessie", "jessie-updates")
 //     site          (eg, "http.debian.net")


mvo5:
Hey @nachoparker ! Thanks for your PR. It is rather unfortunate that there is no distinction between normal updates and security updates in raspbian :/ Given that situation I think its better if unattended-upgrades will just install all updates. What do you think?

a-detiste:
Well, Raspbian is like a two-man-army, they have to cut corners:
- no security tracking
- no unstable, only testing & stable
- "any" non-free, contrib packages, only the "all" get imported
- some packages like UU get patched like in ubuntu; but then get stuck for a long time with an old version, because they need manual intervention
- no i386 build, obviously, but this could be nice to be able to run steamcmd with qemu on Raspbian

That's why I got this template merged in at the first time: to lessen maintainer burden and have smoother updates of UU in RaspBian; this would be so nice to have RaspBian maintainer making this even better.

nachoparker:
@mvo5 yes, it's really unfortunate. I necro'ed that thread to try to catch their attention on this. I am the developer of [NextCloudPi](https://ownyourbits.com/2017/02/13/nextcloud-ready-raspberry-pi-image/) so it is a real bummer for me and my growing user base.

In any case, updating everything will have to do. It's better than leaving the thing insecure.

@a-detiste I know. In case you are interested, I have a wrapper launcher for QEMU with network access so that `apt` works right away ;)

[link](https://github.com/nachoparker/qemu-raspbian-network)

nachoparker:
I develop on top of Raspbian, so I can keep you guys up to date with how things go with UU. Unfortunately it is on the Raspbian devs to implement the Raspbian-Security labels. Nothing I can do about that.

PR #68, opened 2017-07-28
document and enable autoclean feature

anarcat:
without this, unattended-upgrades will slowly fill up /var unless a
sysadmin sets up such an automated process or cleans up those packages
manually.

this is especially important on embeded systems (e.g. raspbian) where
disk space is sparse, but will eventually fill up any hard disk out
there. after a few months of operations here, I had a 6GB
/var/cache/apt/archives directory that eventually blocked email
delivery and caused a downtime.

### DIFF
--- a/data/50unattended-upgrades.Debian
+++ b/data/50unattended-upgrades.Debian
@@ -100,3 +100,5 @@ Unattended-Upgrade::Package-Blacklist {
 // Specify syslog facility. Default is daemon
 // Unattended-Upgrade::SyslogFacility "daemon";
 
+// Do "apt-get autoclean" every n-days (0=disable)
+APT::Periodic::AutocleanInterval "7";
--- a/data/50unattended-upgrades.Raspbian
+++ b/data/50unattended-upgrades.Raspbian
@@ -89,3 +89,6 @@ Unattended-Upgrade::Package-Blacklist {
 // Use apt bandwidth limit feature, this example limits the download
 // speed to 70kb/sec
 //Acquire::http::Dl-Limit "70";
+
+// Do "apt-get autoclean" every n-days (0=disable)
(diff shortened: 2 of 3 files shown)


a-detiste:
on a system with a fragile flash memory + lot of ram you may want to mount /var/cache/apt/archives on a tmpfs

anarcat:
which will still fill up. in any case, this should at least be documented somehow.

mvo5:
Thanks for the PR! I agree with the goal, I'm slightly hesitant about the chosen implementation. This is an apt setting so it feels a bit odd that we set it unconditionally (a user may have already set it differently).

There is a `Unattended-Upgrade::Keep-Debs-After-Install` setting in u-u - it will only remove the debs that u-u touched and only when things were successful. We should simply set to "true" by default (see line 1608 in the code) and document in the config. Do you think you could do that? Happy to merge the PR with that change :)

anarcat:
hmm.. then there's a bug in that option is what is going on here. `Unattended-Upgrade::Keep-Debs-After-Install` defaults to `false`, which, from what i understand, would have cleared the old .deb files from the local cache. but it didn't.

setting it to `true` would *keep* those files, no?

mvo5:
@anarcat Sorry for my slow reply - I just tested this on my ubuntu machine and it seems like the default to clean the debs works for me. I.e. it upgraded e.g. firefox for me but no firefox deb is leftover in /var/cache/apt/archives (and I also add some debug messages to double check). So maybe the filling-up happend for other reasons (or because the version of u-u did not have the feature yet). Keen to figure out the bottom of it, maybe APT::Periodic::Download-Upgradeable-Packages is set?

anarcat:
okay, maybe this is an unrelated issue then. closing.

PR #69, opened 2017-07-30
Increase timeout to 30 minutes for u-u-s to finish before killing it

rbalint:
Soft timeout until u-u-s is considered to be finished properly
is also increased to 25 minutes from 10 minutes.


rbalint:
Switching to minimal steps by default ( https://github.com/mvo5/unattended-upgrades/pull/70 )makes time-outing on upgrades less likely, but increasing the timeout may still make sense.

PR #77, opened 2017-09-19
~90 % speedup and minor fixes

rbalint:


test/test_in_chroot.py
```diff
@@ -85,17 +85,17 @@ def _unpack_debootstrap_tarball(self, tarball, target):
...
         options = MockOptions()
-        options.minimal_upgrade_steps = False
+        options.minimal_upgrade_steps = True
         # run it
         target = self._run_upgrade_test_in_real_chroot(options)
         # ensure we upgraded the expected packages
         self.assertTrue(
             self._verify_install_log_in_real_chroot(target, "ca-certificates"))
 
     def test_minimal_steps_upgrade(self):
```
mvo5: Maybe worth renaming this test as well - like test_not_minimal_steps or something?

rbalint: Ah, indeed!

Issue #32, opened 2016-09-26
raspbian configuration file fails and is not used anyways


anarcat:
Hi,

I am trying to work on bug [LP#1375919](https://bugs.launchpad.net/raspbian/+bug/1375919) (Raspbian: unattended-upgrades does not work in default installation) and I noticed that the [config file shipped here](https://github.com/mvo5/unattended-upgrades/blob/master/data/50unattended-upgrades.Raspbian) does not match the one shipped by raspbian.

Therefore, this creates problems when upgrading raspbian. Even if the files would match, neither file properly allows unattended-upgrades to actually work correctly. The LP bug report mentions the following line work:

```
"origin=Raspbian,codename=jessie,component=main";
```

I can confirm that works. However, it also says that removing the line should allow upgrades, but I wasn't able to confirm that.

It would be nice to have better consistency here, so I figured I would file an issue here as well to bring your attention to this more obscure platform.

anarcat:
Here's the default policy config on raspbian-lite, btw:

```
 500 http://archive.raspberrypi.org/debian/ jessie/ui armhf Packages
     release o=Raspberry Pi Foundation,a=stable,n=jessie,l=Raspberry Pi Foundation,c=ui
     origin archive.raspberrypi.org
 500 http://archive.raspberrypi.org/debian/ jessie/main armhf Packages
     release o=Raspberry Pi Foundation,a=stable,n=jessie,l=Raspberry Pi Foundation,c=main
     origin archive.raspberrypi.org
 500 http://mirrordirector.raspbian.org/raspbian/ jessie/rpi armhf Packages
     release o=Raspbian,a=stable,n=jessie,l=Raspbian,c=rpi
     origin mirrordirector.raspbian.org
 500 http://mirrordirector.raspbian.org/raspbian/ jessie/non-free armhf Packages
     release o=Raspbian,a=stable,n=jessie,l=Raspbian,c=non-free
     origin mirrordirector.raspbian.org
 500 http://mirrordirector.raspbian.org/raspbian/ jessie/contrib armhf Packages
     release o=Raspbian,a=stable,n=jessie,l=Raspbian,c=contrib
     origin mirrordirector.raspbian.org
 500 http://mirrordirector.raspbian.org/raspbian/ jessie/main armhf Packages
     release o=Raspbian,a=stable,n=jessie,l=Raspbian,c=main
     origin mirrordirector.raspbian.org
```

rbalint:
[...]

PR #81, opened 2017-11-09
Fix Raspbian default config codename matching

peternowee:
From my commit message:
> Using `archive=${distro_codename}` does not match, at least not in Raspbian Stretch anymore. Changing it to `codename=${distro_codename}`. I tested that it now works and did not notice any regressions.
> 
> Same change was made to the Debian-specific configuration 3 years ago, see commit 9a2afa5.

Note that this does not solve the current situation of Raspbian distributing the Debian-specific config file instead of the Raspbian-specific config file. This should be solved downstream at Raspbian, as discussed in issue #32. However, once that issue is solved, this change will be necessary anyway.


mvo5:
\o/

peternowee:
Thanks!

PR #83, opened 2017-12-14
Multiple fixes for issues reported to BTS and Launchpad

rbalint:


debian/tests/upgrade-between-snapshots
```diff
@@ -16,7 +16,8 @@ case "$(lsb_release -s -i)" in
         # install u-u to pull in dependencies and mark python3.5-minimal
         # because otherwise u-u autoremoves it while running python3.5 code
         chroot $chroot_dir apt-get install -y unattended-upgrades python3.5-minimal
-        # TODO add a few interesting packages to upgrade
+        # add package set with many dependencies
+        chroot $chroot_dir apt-get install -y xfce4
```
rbalint: This makes the test fail exposing https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=875383

debian/tests/upgrade-between-snapshots
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
+
+set -e
+
+chroot_dir=$AUTOPKGTEST_TMP/chroot
```
mvo5: Nice test!

unattended-upgrade
```diff
@@ -504,15 +504,39 @@ def upgrade_in_minimal_steps(cache,            # type: apt.Cache
     # double check any changes we do
     allowed_origins = get_allowed_origins()
 
+    # pre-calculate set sizes to process sets which are expected to be smaller
+    # earlier
+    upgrade_set_sizes = {}
```
mvo5: Won't this need a mypy type annotation?

unattended-upgrade
```diff
@@ -721,10 +745,16 @@ def conffile_prompt(destFile, prefix=""):
...
     # does not have conffiles anymore
     pkg_conffiles = ""
-    deb = apt_inst.DebFile(destFile)
     try:
+        deb = apt_inst.DebFile(destFile)
         pkg_conffiles = deb.control.extractdata("conffiles").strip().decode(
             "utf-8")
+    except SystemError as e:
+        print(_("Apt returned an error, exiting"))
+        print(_("error message: '%s'") % e)
```
mvo5: Maybe remove this print and just `raise` again at the `sys.exit(1)` point?

unattended-upgrade
```diff
@@ -759,6 +789,15 @@ def conffile_prompt(destFile, prefix=""):
         # record for later
         dpkg_status_conffiles[conf_file] = md5
 
+        # the package replaces a directory wih a configuration file
```
mvo5: Nice catch

unattended-upgrade
rbalint: I have not used mypy on this project but now trying it it reports several issues which I'll resolve in the next PR.
I'm also adding an annotation for this variable.

unattended-upgrade
rbalint: For the users raising the issue would be as informative as printing it, but I'd like to avoid reporting known and handled errors to errors.ubuntu.com.

unattended-upgrade
mvo5: Thanks! I added mypy hints a while ago, I thought I also added a corresponding test that is run in travis. Let me check what is going on there.

PR #84, opened 2017-12-17
Add Devuan support.

evilham:
Dear Michael,

this PR adds the `50unattended-upgrades.Devuan` file, that adds support for Devuan.

Hopefully this can land at some point in Buster (no rush) and we can help you out in the future with bug fixes and the like :).

Best regards,
Evilham


evilham:
That was quick, thanks!

PR #87, opened 2018-01-09
Minor changes

rbalint:


debian/tests/upgrade-between-snapshots
```diff
@@ -5,33 +5,36 @@ set -e
...
+# because otherwise u-u autoremoves it while running python3.5 code
+$chroot_exec apt-get install -y unattended-upgrades python3.5-minimal mailutils
+# add package set with many dependencies
+$chroot_exec apt-get install -y xfce4
+$chroot_exec apt-get clean
+$chroot_exec apt-mark showmanual > $chroot_dir/tmp/manual
+
+# install package version just built
+cp ../unattended-upgrades_*.deb $chroot_dir/tmp/
+$chroot_exec bash -c 'dpkg -i /tmp/unattended-upgrades_*deb' 2>&1
```
mvo5: (nitpick) we could use `apt install -y /tmp/unattended-upgrades_*.deb` here now (unless we care about running this on old distros.

unattended-upgrade-shutdown
```diff
@@ -74,7 +74,7 @@ def log_progress():
     """ helper to log the install progress (if any) """
     # wait a some seconds and try again
     msg = _("Unattended-upgrade in progress during shutdown, "
-            "sleeping for 5s")
+            "please don't turn off the computer")
```
mvo5: \o/

debian/tests/upgrade-between-snapshots
```diff
@@ -7,32 +7,34 @@ start_date=20171001T000000Z
...
+
+# install Debian's configuration file to make this test work as expected even on other distributions
+cp data/50unattended-upgrades.Debian /etc/apt/apt.conf.d/
```
rbalint: This will not work on Ubuntu, but I already fixed it just forgot squashing the commit to this one.

debian/tests/upgrade-between-snapshots
rbalint: I fixed that in an additional commit in the PR. It is indeed a bin nicer.

mvo5:
Looks good, thanks for addressing my comments! Is the conffile in the autopkgtest  good (looks like its always using the debian one)? if so, feel free to merge.

rbalint:
2018-01-10 18:49 GMT+01:00 Michael Vogt <notifications@github.com>:

> Looks good, thanks for addressing my comments! Is the conffile in the
> autopkgtest good (looks like its always using the debian one)? if so, feel
> free to merge.
>
Yes, since we always test the snapshots of Debian and we need a matching
origin patterns. When Ubuntu creates a snapshot service I'll be more than
happy to switch to it when testing u-u on Ubuntu.

PR #91, opened 2018-02-08
Last set ot changes for 0.99 and finalizing for upload

rbalint:


unattended-upgrade
```diff
@@ -472,6 +472,7 @@ def cache_commit(cache,           # type: apt.Cache
     with Unlocked():
         try:
             res = cache.commit(install_progress=iprogress)
+            cache.open()
```
mvo5: Nice catch!

PR #92, opened 2018-02-10
Consider conffile prompts to be errors

### LINKED ISSUE #89: Does not send email when package has conffile prompt and needs manual upgrade
Hi,

I am using version 0.93.1+nmu1 on Raspbian Stretch. My version is a bit old but if I am not wrong upstream changelog does not mention a fix for my issue.

I have unattended-upgrades configured to send an email only when an error occurs. However, it does not send an email when an update has a conffile prompt and thus needs manual upgrade. This seems to be an important bug to me, because the administrator believes his system is up to date while it is not.

This was working "before" (see for example 0.86 changelog). I reported this bug with details in [Debian #852465](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=852465). Do not hesitate to ask if I can provide more information or test things.

Thanks for your work,
Yvan

nomis:
Flag packages that have to be upgraded manually because of a conffile
prompt and consider this to be an error when sending email or exiting.

Fixes #89.

### DIFF
--- a/test/test_mail.py
+++ b/test/test_mail.py
@@ -154,15 +154,6 @@ Debian-Security']
         self.assertTrue(
             "Packages that attempted to upgrade:\n 2vcard" in mail_txt)
 
-    def test_mail_on_error_with_warning_in_log(self):
-        apt_pkg.config.set("Unattended-Upgrade::MailOnlyOnError", "true")
-        pkgs, res, pkgs_kept_back, mem_log, logf_dpkg = self._return_mock_data(
-            successful=True)
-        mem_log.write("\nWARNING: some warning\n")
-        send_summary_mail(pkgs, res, pkgs_kept_back, mem_log, logf_dpkg)
-        self.assertTrue(
-            os.path.exists(os.path.join(self.tmpdir, "mail.txt")))
-
     def test_summary_mail_blacklisted(self):
         # Test that blacklisted packages are mentioned in the mail message.
         send_summary_mail(*self._return_mock_data())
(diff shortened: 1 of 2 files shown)


test/test_mail.py
```diff
@@ -154,15 +154,6 @@ def test_summary_mail_only_on_error(self):
         self.assertTrue(
             "Packages that attempted to upgrade:\n 2vcard" in mail_txt)
 
-    def test_mail_on_error_with_warning_in_log(self):
```
mvo5: I have not looked deeper but it appears this change (and the one below with the `re.search()` for the warning revert the fix for https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=703621 - is there an alternative for this? Or was the initial fix bad?

unattended-upgrade
```diff
@@ -1495,6 +1492,8 @@ def main(options, rootdir=""):
     if options.download_only:
         return 0
 
+    pkg_install_manual = False
```
mvo5: I think I would prefer `pkg_has_conffile_prompt` for this

test/test_mail.py
mvo5: Aha, I read up the details about this now and it looks like the initial fix was indeed bad. Thanks for catching this.

unattended-upgrade
```diff
@@ -1664,7 +1664,8 @@ def main(options, rootdir=""):
     if not options.dry_run:
         log_content = get_dpkg_log_content(logfile_dpkg, install_start_time)
         send_summary_mail(
```
mvo5: Given that we use `pkg_install_sucess and pkg_install manual` here and below I think it makes sense to have something like: `successful_run = pkg_install_success and not pkg_install_manual`.

mvo5:
Thanks @nomis !

PR #95, opened 2018-02-13
Amd64 only snapshot upgrade autopkgtest

rbalint:


debian/changelog
```diff
@@ -1,3 +1,9 @@
+unattended-upgrades (1.0) UNRELEASED; urgency=medium
```
mvo5: hm, lets call it 0.100 ?

mvo5:
I cherry-picked fb6e35170e33d0b032747d817eb3d32139aedf14 - how do you feel about generating the changelog via `gbp dch` before the upload?

rbalint:
I prefer opening the changelog before adding changes to avoid generating debs with the same name as uploaded ones but with different content, but I can just skip the changelog updates from the PRs.

debian/changelog
rbalint: @mvo5 is there anything else you would like to see in a release to call it 1.0?

mvo5: @rbalint I was thinking about this a bit and I think I agree - its pretty feature complete (once we deal with kernels better) and I don't think there is any feature missing at this point. So :+1: for 1.0. What are you plans with the kernel? Will that happen before 18.04?

rbalint: @mvo5 If nothing urgent comes up, before weekend. ;-)

mvo5: @rbalint yay!

PR #97, opened 2018-02-15
Remove autoremovable kernels

rbalint:


rbalint:
I tested this on a Xenial system which accumulated a fair amount of kernels and it worked well. Maybe adding a test would be useful, but i'd prefer discussing the approach first to make sure we agree on it.

julian-klode:
I'm not sure what there is to gain from doing the autoremoval in minimal steps. they're safe to remove, so it seems to me best would be to just do them all at once.

Rather than a(nother) fixed list, it seems to make more sense to reuse APT::VersionedKernelPackages.

unattended-upgrade
```diff
@@ -1535,8 +1598,28 @@ def main(options, rootdir=""):
...
+    # get log
+    logfile_dpkg = os.path.join(_get_logdir(), 'unattended-upgrades-dpkg.log')
+    if not os.path.exists(logfile_dpkg):
+        with open(logfile_dpkg, 'w'):
+            pass
+
+    # auto-removals
+    auto_removable = get_auto_removable(cache)
+    if (auto_removable and apt_pkg.config.find_b(
+            "Unattended-Upgrade::Remove-Unused-Kernel-Packages", True)):
```
mvo5: I'm curious, why `True` here but its overriden in all the configs (in the data dir) to `false` ?

mvo5:
@julian-klode I would assume minimal steps so that e.g. on shutdown the amount of operations that needs completion before the shutdown can continue is as small as possible.

rbalint:
2018-02-16 2:14 GMT+07:00 Julian Andres Klode <notifications@github.com>:
>
> I'm not sure what there is to gain from doing the autoremoval in minimal steps. they're safe to remove, so it seems to me best would be to just do them all at once.

I performed autoremoval on an old Trusty laptop which collected dozens
of unused kernel packages and also used a slow disk. The removals took
more than a half hour and the shutdown timer would have expired and
u-u would have been killed if u-u performed this removal.

> Rather than a(nother) fixed list, it seems to make more sense to reuse APT::VersionedKernelPackages.

Thanks, I'm switching to use that list. I will also fix LP: #1698159
with APT getting fixed.
emoval in minimal steps. they're safe to remove, so it seems to me
best would be to just do them all at once.\r\n\r\nRather than
a(nother) fixed list, it seems to make more sense to reuse
APT::VersionedKernelPackages. "}],"action":{"name":"View Pull
Request","url":"https://github.com/mvo5/unattended-upgrades/pull/97#issuecomment-366031311"}}}

unattended-upgrade
rbalint: This seems to be the pattern for options enabled by default. ```if apt_pkg.config.find_b(..., True)``` in the code and ```//... "false"``` in the configs.

mvo5: Indeed it is, sorry. These are the commented out options.

rbalint:
I extended the tests to cover kernel removal a bit, I think the change is now good to go.

unattended-upgrade
```diff
@@ -493,6 +501,16 @@ def upgrade_normal(cache, logfile_dpkg, verbose):
     return res
 
 
+def use_minimal_steps(options):
```
mvo5: Nice, I like this helper!

unattended-upgrade
```diff
@@ -1256,24 +1269,88 @@ def get_auto_removable(cache):
...
-        if not dry_run:
-            logging.debug("marking %s for remove" % pkgname)
+    if minimal_steps:
+        for pkgname in auto_removable:
+            if SIGNAL_STOP_REQUEST:
+                logging.warning("SIGNAL received, stopping")
+                return False
+            logging.debug("marking %s for removal" % pkgname)
+            cache[pkgname].mark_delete()
+            changes = set([p.name for p in cache.get_changes()])
```
mvo5: This looks quite similar to the code below - maybe a small helper can be extracted here?

rbalint: It is similar, but does not seem like a good candidate for abstracting it out and adding a new redirection, but I'm not absolutely sure I got your idea right.

mvo5: I was thinking about something like: http://paste.ubuntu.com/p/3D4Q7rSX6y/ - I understand the generated message is slightly different so maybe its overkill. Or a slightly modified version with a different error message http://paste.ubuntu.com/p/3WsGxyjMXw/ - please let me know if I overlooked something (maybe not enough tea yet ;)

rbalint: Ah, OK, I see now. Feel free to add a commit on top of this.

julian-klode:
lgtm from a quick look

mvo5:
Thanks @rbalint ! I added the small helper extraction in master now. Great work.

rbalint:
Hm, it needs a minor follow up since while actions are performed properly the reported list of removed packages may not be accurate because of possibly skipping a few package removals.

PR #98, opened 2018-02-22
Further changes related to unused kernel packages' removal

rbalint:


rbalint:
... and finalize 1.0.

unattended-upgrade
```diff
@@ -967,6 +997,16 @@ def send_summary_mail(pkgs, res, pkgs_kept_back, mem_log, dpkg_log_content):
         body += " " + wrap(" ".join(pkgs_kept_back), 70, " ")
         body += "\n"
     body += "\n"
+    if pkgs_removed:
+        body += _("Packages that were auto-removed:\n")
```
mvo5: :heart:

unattended-upgrade
```diff
@@ -1265,86 +1305,97 @@ def get_dpkg_log_content(logfile_dpkg, install_start_time):
...
                 _("Keeping %s auto-removable package(s) because it would"
                   " also remove the following packages which should "
                   "be kept in this step: %s"), len(auto_removable),
-                " ".join(sorted(changes - auto_removable)))
+                " ".join(sorted(pkgnames - auto_removable)))
 
         cache.clear()
         return False
+    for packagename in pkgnames:
+        if cache.running_kernel_pkgs_regexp and \
```
mvo5: (nitpick^2): I personally prefer using () instead of \, so: 
```
if (cache.running_kernel... and 
        cache....):
```

unattended-upgrade
```diff
@@ -1265,86 +1305,97 @@ def get_dpkg_log_content(logfile_dpkg, install_start_time):
...
+                  "kernel: %s"), packagename, " ".join(sorted(pkgnames)))
+            cache.clear()
+            return False
     return True
 
 
-def do_auto_remove(cache, auto_removable, logfile_dpkg, minimal_steps,
-                   verbose=False, dry_run=False):
-    # type: (apt.Cache, AbstractSet[str], str, bool, bool, bool) -> bool
+def do_auto_remove(cache,             # type: apt.Cache
```
mvo5: Nice! Curious does the dep484 stuff helps you? I find it quite nice but I wonder if that is just me :)

unattended-upgrade
```diff
@@ -1637,13 +1693,21 @@ def main(options, rootdir=""):
         #
         # this step may also remove _auto-removable_ reverse dependencies
         # of kernel packages
-        auto_removable_kernel_pkgs = only_kernel_pkgs(auto_removable)
+        auto_removable_kernel_pkgs = {
```
mvo5: I would love to see some unit tests around this new feature but I do understand this is a bit tricky.

unattended-upgrade
rbalint: Me too, I converted some lines to use ... \ to avoid indenting the same amount as the next lines

unattended-upgrade
rbalint: Sometimes it catches errors which is good, but I prefer when a language automatic type inference with strong types and we don't have to add redundant comments to the code. All in all it still seems to worth the effort here.

PR #100, opened 2018-03-01
Update .travis.yml to actually build and test u-u from the repo

rbalint:


.travis.yml
```diff
@@ -9,12 +9,14 @@ dist: trusty
...
+ - sudo apt-get install -y dh-systemd fakeroot python-mock python3-mock
  - which python3
  - sudo apt-get install python3-pip
  - sudo python3 -m pip install -U mypy
 # travis is "funny" it has a non-distro python3 that will not find
 # python modules installed via apt-get
 script:
-  - (cd test ; PATH=/usr/bin:/usr/local/bin:$PATH make)
-  - sudo unattended-upgrades -v
+  - dpkg-buildpackage -d -us -uc -rfakeroot
```
mvo5: Looks great! Maybe a small comment like `# note that tests are run during package build`  ?

rbalint: done

.travis.yml
```diff
@@ -9,12 +9,16 @@ dist: trusty
...
 # travis is "funny" it has a non-distro python3 that will not find
 # python modules installed via apt-get
+# note that tests are run during package build
+# the full upgrade test is failing on Travis, but autopkgtest covers that, too
 script:
-  - (cd test ; PATH=/usr/bin:/usr/local/bin:$PATH make)
-  - sudo unattended-upgrades -v
+  - dpkg-buildpackage -d -us -uc -rfakeroot
+  - sudo dpkg -i ../unattended-upgrades_*.deb
+  - (sudo unattended-upgrades -v --debug | grep -v '^matching .* against ' || true)
```
mvo5: I would probably just remove this or make it a smoke test like `unattended-upgrades --help` works or something. If we don't get anything reliable from that run, why run it at all?

rbalint: OK, let it be an install test, then just showing help.

rbalint: Pushed, this makes the Travis test much faster at least.

PR #106, opened 2018-03-13
d/rules: Exclude mypy cache from source package

smcv:
The 1.0 release to Debian included this cache, which doesn't seem to be source code.


rbalint:
Thanks!

PR #109, opened 2018-04-12
Avoid crash due to using python-apt's API the wrong way

rbalint:


unattended-upgrade
```diff
@@ -149,13 +149,15 @@ class UnattendedUpgradesCache(apt.Cache):
         # type: (apt.progress.base.Progress) -> None
         apt.Cache.open(self, progress)
         # ensure we update the candidate versions
+        self.candidates_to_adjust = self.get_candidates_to_adjust()
```
mvo5: Thanks for finding/fixing this bug! Great work! I looked at this code and it seems we can go a little bit further with the fixing/refactor. It seems prudent to extract a small "adjust_candidates()" helper that does the for loop. I also would prefer to not use self.candidates_to_adjust anymore because its just too easy to reuse the broken apt.Version iter again. Instead it looks like we just need to cache the list of pkgnames. I drafted what I have in mind here: http://paste.ubuntu.com/p/7wVRcxBmZn/ - it is not mypy clean yet and self._cached_candidate_pkgnames is a name I'm not super happy with. But I hope it gives the idea. Plus I think it needs a bit of documentation that the apt.Version should never reused. WDYT?

rbalint: @mvo5 I like the idea but I would like to revisit this after Bionic's release.

mvo5: Sure thing!

PR #110, opened 2018-04-16
Mutiple fixes and new release

rbalint:


unattended-upgrade
```diff
@@ -1380,8 +1380,13 @@ def do_auto_remove(cache,             # type: apt.Cache
                 logging.warning("SIGNAL received, stopping")
                 pkgs_kept_installed = list(auto_removable - set(pkgs_removed))
                 return (False, pkgs_removed, pkgs_kept_installed)
-            logging.debug("marking %s for removal" % pkgname)
-            cache[pkgname].mark_delete()
+            try:
```
rbalint: Let me fix this in a different way

rbalint: Now it is better IMO.

PR #111, opened 2018-05-16
unattended-upgrades: fix Unlocked context manager

mvo5:
The Unlocked context manager did correctly unlock but did not
reacquire the lock which means that in minimal-upgrade step
mode it is possible to run apt code without a lock. If something
else (like landscape, apt, synaptic, packagekit) locks the cache
in the meantime this will work and u-u will get dpkg errors
because dpkg will not be able to perform its operations. It is
less of an issue in non-minimal mode, but even then the auto-remove
step may fail in this way.

We should also SRU this fix.

### DIFF
--- a/unattended-upgrade
+++ b/unattended-upgrade
@@ -348,7 +348,7 @@ class Unlocked:
     def __exit__(self, exc_type, exc_value, exc_tb):
         # type: (object, object, object) -> None
         try:
-            apt_pkg.pkgsystem_unlock()
+            apt_pkg.pkgsystem_lock()
         except Exception:
             pass
 


unattended-upgrade
```diff
@@ -348,7 +348,7 @@ class Unlocked:
     def __exit__(self, exc_type, exc_value, exc_tb):
         # type: (object, object, object) -> None
         try:
-            apt_pkg.pkgsystem_unlock()
+            apt_pkg.pkgsystem_lock()
```
rbalint: Nice catch!

julian-klode:
It's still broken, as python-apt is broken. IMO, we must never unlock in client applications, and hold commit() while holding the lock - as 

1. that lock will become the frontend lock. That should _never_ be released
2. the lock must be held while locking archives

But `apt.Cache.commit()`, in contrast to `apt_pkg.DepCache.commit()`, requires being called unlocked atm. Not sure how to fix this optimally.

mvo5:
I suspect the underlying issue is that apt needs a more global lock - I guess this is what you mean with frontend-lock? Right now it is using the dpkg lock but there is a race because apt needs to release that in order to let dpkg do the work. so a higher level /var/lib/apt/lock would be in order that everything libapt respects. this can then be kept being locked even when the lower-level dpkg lock is released.

julian-klode:
So, system lock actually is supposed to manage the frontend lock soon, and there's a separate method to just release the normal dpkg lock, as my intention was that you can do:

```
with apt_pkg.SystemLock():
 ...
 cache.commit()
 ...
```

and everything would just work. But since the `apt` module is broken wrt `commit()`, we probably need changes everywhere.

rbalint:
@mvo5 @julian-klode Even if locking gets fixed in python-apt API imo this fix should go in, i just created https://github.com/mvo5/unattended-upgrades/pull/112 to fix tests first.

rbalint:
Merging for now, it makes the situation much better and later we can adopt the python-apt API changes, too.

PR #114, opened 2018-05-23
Multiple fixes and releasing 1.2

rbalint:


debian/changelog
```diff
@@ -1,3 +1,27 @@
...
+    in the meantime this will work and u-u will get dpkg errors
+    because dpkg will not be able to perform its operations. It is
+    less of an issue in non-minimal mode, but even then the auto-remove
+    step may fail in this way.
+
+  [ Balint Reczey ]
+  * Run mypy with --no-strict-optional to keep mypy 0.600 happy.
+    Older versions does not support Optional[] typing, thus adding it to the
+    comments would break the build with all old releases.
+  * End Python 2 support.
```
mvo5: This makes me slightly sad but I guess nowdays there is really no distro anymore that needs python2/python-apt so thats fine. The goal with py2 support was always to make sure that older distros could use a backported u-u.

test/test_pep484.py
```diff
@@ -22,7 +22,8 @@ def test_pep484_clean(self):
         # --ignore-missing-imports
         self.assertEqual(
             subprocess.call(
-                ["mypy", "--ignore-missing-imports", top_src_dir]), 0)
+                ["mypy", "--ignore-missing-imports",
+                 "--no-strict-optional", top_src_dir]), 0)
```
mvo5: Thanks for this!

unattended-upgrade
```diff
@@ -566,7 +563,7 @@ def upgrade_in_minimal_steps(cache,            # type: apt.Cache
             pkg.mark_install()
         else:
             continue
-        upgrade_set_sizes[pkgname] = len(cache.get_changes())
+        upgrade_set_sizes[pkgname] = cache.install_count
```
mvo5: nice

PR #115, opened 2018-05-24
Add logging for scheduled host reboot

irregulator:
The shutdown command will print scheduled reboot to standard
error. This is not reflected in unattended-upgrades logs. What
is more, on Debian systems unattended-upgrade is typically invoked
by apt daily cron. Printing in stderr will result in cron email
notification to root user, which might be undesired given this
is not an actual error.

This commit grabs the reboot message and logs it. In case of
failure, exception will be logged as well.

test_reboot tests are modified accordingly. Since
reboot_if_requested_and_needed() calls logged_in_users() and both
end up using subprocess it's better to decouple them during tests
and avoid uncontrolled behavior. This is achieved by mocking
logged_in_users() and manipulating its return value.

### DIFF
--- a/test/test_reboot.py
+++ b/test/test_reboot.py
@@ -4,6 +4,7 @@ import datetime
 import logging
 import os
 import unittest
+import subprocess
 
 import apt_pkg
 apt_pkg.config.set("Dir", "./aptroot")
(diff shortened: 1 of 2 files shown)


irregulator:
CI failed with
```
unattended_upgrade.py:1211: error: Module has no attribute "run"
../test/unattended_upgrade.py:1211: error: Module has no attribute "run"
F
======================================================================
FAIL: test_pep484_clean (__main__.PackagePep484TestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "./test_pep484.py", line 26, in test_pep484_clean
    "--no-strict-optional", top_src_dir]), 0)
AssertionError: 1 != 0
----------------------------------------------------------------------
Ran 1 test in 4.220s
```

Alas subprocess.run was introduced in python 3.5 but CI is running Ubuntu Trusty with python 3.4 . 

Do we target such systems for new features? I guess I have to rework the patch with subprocces.check_output or similar.

rbalint:
@julian-klode plans changing the Travis tests to be run in a bionic docker env which will solve this failure. IMO it is enugh to target bionic and more recent systems.

mvo5:
If its not too much hassle I would love if this would still run on trusty. I know people backport u-u to older releases, especially now that @rbalint and friends are so active with features and fixes. So if its not too much of a burden I would love if we keep this easy for people to do.

c00kiemon5ter:
I tend to support anything that latest Debian Stable supports, where the current python3 version is `3.5.3`. As such I would like to see projects move there, and not cling to legacy platform mess.

Moreover, is Ubuntu Trusty (as an LTS release) going to backport feature updates? or just security fixes? There is no reason to be constrained by older systems, if newer features are not going to run on them.

rbalint:
@mvo5 @c00kiemon5ter I do plan back-porting many changes to Trusty, but IMO this one is not badly needed and leaving it as it is fine.

mvo5:
This is now merged via https://github.com/mvo5/unattended-upgrades/pull/142 - thanks for your PR and sorry that it took so long to get merged :/ If there is anything you dislike about my changes or that does not work as you expected, please do let me know.

Issue #116, opened 2018-06-05
Blacklisted packages are included in the "upgradable origin", while they should not


zerkms:
I'm running the following config on bionic:

```
Unattended-Upgrade::Allowed-Origins {
        "${distro_id}:${distro_codename}-security";
};

Unattended-Upgrade::Package-Blacklist {
        "haproxy";
        "linux-aws";
        "linux-virtual";
        "linux-generic";
        "mongodb";
        "nginx";
        "pgbouncer";
        "php.*fpm";
        "postgresql";
        "rabbitmq";
        "redis";
        "supervisor";
};
```

So only `bionic-security` origin should be checked when unattended upgrades are considered.

Few days ago a new `postgresql-10` package version was pushed to the `bionic-updates` yet when I run `unattended-upgrades` it is misleadingly included in the email:

```
Unattended upgrade returned: True

Packages that were upgraded:
 
Packages with upgradable origin but kept back:
 postgresql-10 postgresql-client-10 



Unattended-upgrades log:
Initial blacklisted packages: haproxy linux-aws linux-virtual linux-generic mongodb nginx pgbouncer php.*fpm postgresql rabbitmq redis supervisor
Initial whitelisted packages: 
Starting unattended upgrades script
Allowed origins are: o=Ubuntu,a=bionic-security
Using (^linux-image|^linux-headers|^linux-image-extra|^linux-modules|^linux-modules-extra|^linux-signed-image|^kfreebsd-image|^kfreebsd-headers|^gnumach-image|^.*-modules|^.*-kernel|^linux-backports-modules-.*|^linux-modules-.*|^linux-tools|^linux-cloud-tools) regexp to find kernel packages
[...]

PR #117, opened 2018-06-05
Changed the order of origin and blacklisting checks

zerkms:
This is a suggested fix for #116 

Unfortunately recreating the complete state to of the repository index looks too complicated for me at the moment, so I cannot provide a test for it, sorry :-(


zerkms:
@rbalint I agree, see the improved fix

zerkms:
@rbalint I removed side effects from `is_pkgname_in_blacklist` and now it's indeed cleaner and better solution

rbalint:
@zerkms Thanks! Could you please squash the two commits? It helps git bisecting when all the commits are passing tests.
Other tests also divert aptroot to not be affected by system's apt configuration and doing it here, too, would also be cleaner.

zerkms:
@rbalint totally, now it's just one commit

PR #118, opened 2018-06-06
Pu/types fixes

julian-klode:
Adjust types to work with upcoming python-apt type hints. There is still one failure:

```test/unattended_upgrade.py:1147: error: Argument "facility" to "SysLogHandler" has incompatible type "str"; expected "int"```

Which seems correct according to documentation, but that might be a bug in documentation and mypy's type hints.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,10 +14,11 @@ before_install:
  - which python3
  - sudo apt-get install python3-pip
  - sudo python3 -m pip install -U mypy coverage
+ - git clone https://salsa.debian.org/apt-team/python-apt ../python-apt
 # travis is "funny" it has a non-distro python3 that will not find
 # python modules installed via apt-get
 # note that tests are run during package build
 script:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by julian-klode** [comment]
I'll add a commit soon to work around the syslog handler issue and enable checking against the python-apt git repo. This allows us to check against the latest type annotations rather than stale ones which might contain bugs.

**Comment by mvo5** [comment]
Looks great! There is a test failure though:
```
OK
Running ./test_pep484.py with python3
mypy 0.600
/home/travis/build/mvo5/python-apt/apt/progress/text.py:122: error: Name 'signal.Signals' is not defined
/home/travis/build/mvo5/python-apt/apt/progress/text.py:122: error: Name 'signal.Handlers' is not defined
```
which appears to be coming from the python-apt from git?

julian-klode:
Yup, I think I'll update the pull request to run the CI test in a recent bionic via docker, then we get an up-tod-date python.

PR #119, opened 2018-06-06
Fix crashing while adjusting candidates

rbalint:
LP: #1775307


rbalint:
It was shadowed by https://github.com/mvo5/unattended-upgrades/pull/114/commits/b2518a17c4a53a5f3c165c36ad9e251fa56e87db and also by not having tests where candidate adjustment is needed.

PR #120, opened 2018-06-06
Ignored `.idea` and `.mypy_cache` directories

zerkms:
Under no circumstances any of those would be committed ever.

How about ignoring them so that `git status` output was cleaner?


zerkms:
@rbalint while you're here - do I need to do something additional at https://github.com/mvo5/unattended-upgrades/pull/117 ?

PR #123, opened 2018-06-13
WIP Checked locking

rbalint:
This fails tests at the moment because python-apt loses the lock

### DIFF
--- a/unattended-upgrade
+++ b/unattended-upgrade
@@ -1509,6 +1509,15 @@ def main(options, rootdir=""):
         logging.error("Lock file is already taken, exiting")
         sys.exit(1)
 
+    # check and get system lock (/var/lib/dpkg/lock)
+    try:
+        apt_pkg.pkgsystem_lock()
+    except SystemError as e:
+        logging.error(_("Lock could not be acquired (another package "
+                        "manager running?)"))
+        print(_("Cache lock can not be acquired, exiting"))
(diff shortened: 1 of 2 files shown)


unattended-upgrade
```diff
@@ -336,22 +336,22 @@ class LogInstallProgress(apt.progress.base.InstallProgress):
...
+    Context manager for unlocking the apt lock for cache.commit() or dpkg call
     """
 
     def __enter__(self):
         # type: () -> None
         try:
             apt_pkg.pkgsystem_unlock()
         except Exception:
-            pass
+            raise
```
rbalint: @mvo5 This is why we missed locking problems earlier. :-(

rbalint:
When python-apt is fixed I'll add the versioned dependency

unattended-upgrade
mvo5: Thanks for finding this!

mvo5: Given that try/except now just raises, we might as well remove it, no?

rbalint: True, fixed.
Originally I wanted to add some logging but later decided it would have been redundant.

rbalint:
superseded by https://github.com/mvo5/unattended-upgrades/pull/129 which converts the exception on unlock to a warning to let it work with buggy python-apt versions

PR #128, opened 2018-06-23
Test installing security updates on stable releases and other minor changes

rbalint:


unattended-upgrade
```diff
@@ -388,13 +391,29 @@ def should_stop():
...
+    global MISSING_NMCLI_PRINTED
     if SIGNAL_STOP_REQUEST:
         logging.warning("SIGNAL received, stopping")
         return True
     if apt_pkg.config.find_b("Unattended-Upgrade::OnlyOnACPower", True) and \
        subprocess.call("on_ac_power") == 1:
         logging.warning("System is on battery power, stopping")
         return True
+    if apt_pkg.config.find_b(
+            "Unattended-Upgrade::Skip-Updates-On-Metered-Connections", True):
```
rbalint: @julian-klode apt could use this config option to prevent downloads for u-u, too.

debian/control
```diff
@@ -17,17 +18,19 @@ Build-Depends-Indep: python3-dev,
...
-Standards-Version: 3.9.3
+Standards-Version: 4.1.4
 Vcs-Git: https://github.com/mvo5/unattended-upgrades.git
 
 Package: unattended-upgrades
 Architecture: all
 Depends: ${shlibs:Depends},
          ${misc:Depends},
          debconf,
+         gir1.2-glib-2.0,
```
mvo5: Can we make those dependencies optional please? And handle them missing gracefully via import errors and warnings?

rbalint: The packages are present even in minimal lxc containers.

mvo5: It's fine, if it bothers me too much I can just make a PR to fix it - its just my instinct to avoid the extra 3 packages (and ~1MB) for a (admittedly nice) feature that is off by default (especially since python handles missing imports so nicely).

debian/tests/common-functions
```diff
@@ -0,0 +1,73 @@
+#!/bin/sh
+
```
mvo5: Probably a good idea to run "shellcheck" over this script (and ideally make it part of the unit tests to run shellcheck on this and all other shell scripts).

debian/tests/common-functions
```diff
@@ -0,0 +1,73 @@
...
+
+check_arch() {
+    local architecture
+    architecture=$(dpkg --print-architecture)
```
mvo5: Not strictly needed here but I would still suggest to use quotes around all "$" calls just for good measure.

debian/tests/common-functions
```diff
@@ -0,0 +1,73 @@
...
+        *)
+            echo "Skipping upgrade test on $architecture"
+            exit 0
+            ;;
+    esac
+}
+
+chroot_exec() {
+    local chroot_dir
+    chroot_dir=$1
```
mvo5: Quotes here will avoid breakage if the dir ever have spaces.

debian/tests/upgrade-all-security
```diff
@@ -0,0 +1,61 @@
...
+# On development releases the latest stable release is tested since there are no
+# security updates to test with.
+
+set -e
+
+chroot_dir=$AUTOPKGTEST_TMP/chroot
+
+distro=$(lsb_release -c -s)
+# fall back to latest release on any development release
+if [ $distro = sid ] || lsb_release -d -s | grep -q 'development branch'; then
```
mvo5: Distro should be quoted.

debian/tests/upgrade-all-security
```diff
@@ -0,0 +1,61 @@
...
+# save list of manually installed packages
+chroot_exec $chroot_dir apt-mark showmanual > $chroot_dir/tmp/manual
+
+# clean up to need less space for the test
+chroot_exec $chroot_dir apt-get clean
+
+run_u_u $chroot_dir
+
+echo "Checking if there is anything left not upgraded:"
+chroot_exec $chroot_dir apt-get upgrade -s | tee $chroot_dir/tmp/updates-left
```
mvo5: I would suggest to use `apt upgrade -s` here - apt-get upgrade may held back packages that require new dependencies. This means that e.g. on ubuntu a new kernel will make this test fail. Far less common on Debian but *might* happen AFAIK.

debian/control
rbalint: I'm trying to avoid extra dependencies, too, unless they are present anyway practically everywhere.

debian/tests/common-functions
rbalint: It is already run by check-all-the-things, thus adding it per package feels redundant.

rbalint: But thanks, I used it to check the scripts.

mvo5: Aha, thank you! I was not aware that is already used, nice!

debian/tests/upgrade-all-security
```diff
@@ -0,0 +1,61 @@
...
+# save list of manually installed packages
+chroot_exec "$chroot_dir" apt-mark showmanual > "$chroot_dir/tmp/manual"
+
+# clean up to need less space for the test
+chroot_exec "$chroot_dir" apt-get clean
+
+run_u_u "$chroot_dir"
+
+echo "Checking if there is anything left not upgraded:"
+chroot_exec "$chroot_dir" apt-get upgrade --with-new-pkgs -s | tee "$chroot_dir/tmp/updates-left"
```
mvo5: :+1:

PR #129, opened 2018-06-25
Unlock for dpkg operations with apt_pkg.pkgsystem_unlock_inner() 

rbalint:
… when it is available

Also stop running when reacquiring the lock fails.


rbalint:
@julian-klode This change is intended to make u-u work with and without frontend locking per https://lists.ubuntu.com/archives/ubuntu-devel/2018-June/040381.html

PR #132, opened 2018-06-29
Use fully qualified domain name in email subject

### LINKED ISSUE #107: Allow mail subject personalization
When receiving mail from nodes which have the same host name but have different domain names (e.g. *foo.example.com* and *foo.bar.example.net*), the subject of the message only contains the host part of the FQDN (in this example `Subject: unattended-upgrades result for 'foo'`) which makes it harder to distinguish these hosts.

While you can refer to the sender to find out which node has sent the message, it might be convenient for the user to have the FQDN in the mail title, or maybe even personalize this title with some configuration settings (i.e. when the config is generated by configuration management tools, the subject could include a customer ID).

Do you have an opinion about this?

rbalint:
Fixes #107.

### DIFF
--- a/unattended-upgrade
+++ b/unattended-upgrade
@@ -38,6 +38,7 @@ import re
 import os
 import select
 import signal
+import socket
 import string
 import subprocess
 import sys
@@ -909,7 +910,7 @@ def rewind_cache(cache, pkgs_to_upgrade):
 
 def host():
     # type: () -> str
-    return os.uname()[1]
+    return socket.getfqdn()
 
 
 # *sigh* textwrap is nice, but it breaks "linux-image" into two


rbalint:
@mvo5 Autopkgtest output shows the changed result but on my system it just gets a hostname in qemu. At least it is easy to check that there is no regression.

uxmaster:
@rbalint, actually, there's a regression since `socket.getfqdn()` always returns `localhost` instead of the hostname (set in `/etc/hostname`) for all my servers.

To work around this, I had to set the real hostname in `/etc/hosts` as `127.0.1.1` and `::1`.

rbalint:
@uxmaster That does not sound like a workaround, but the proper configuration.

uxmaster:
You're probably right, but the default configuration of Ubuntu at Vultr and Google Cloud seems not to have these defined properly. While at Google the `socket.getfqdn()` returns what the reverse DNS says about LAN IP, but at Vultr it defaults to `localhost`. Maybe it is not the current default, but a leftover from previous Ubuntu releases.

s0ftice:
not a big fan of this "fix".
my subject is now
`unattended-upgrades result for ec2-xxx-xxx-xxx-xxx.ap-southeast-1.compute.amazonaws.com`
and that for multiple servers, so I don't know which server a certain notification is about (on first look).

This misleading FQDN is not even configured in my /etc/hostname or /etc/hosts.
A `grep -R` also does not find ec2-xxx-xxx-xxx-xxx anywhere in /etc.
Looks like an reverse DNS entry, which I can only discover by running `hostname -A` on my (upgraded) Debian 11.

I would actually like to have the output of `hostname -fqdn` in the subject line, which is my proper FQDN (might not be `socket.gethostname()`.)

PR #133, opened 2018-07-03
Send emails on errors or exceptions occured

rbalint:
This helps monitoring hosts running u-u and also helps getting bug reports from Debian users because the exceptions occurring on Debian are not collected at errors.ubuntu.com.

### DIFF
--- a/test/test_mail.py
+++ b/test/test_mail.py
@@ -4,6 +4,8 @@ from __future__ import unicode_literals
 
 from email.parser import Parser
 from io import StringIO
+import logging
+from mock import patch
 import os
 import shutil
 import sys
@@ -24,11 +26,21 @@ from unattended_upgrade import (
 )
 
 
+class MockOptions(object):
+    debug = False
+    verbose = False
+    download_only = False
+    dry_run = False
+    apt_debug = False
+    minimal_upgrade_steps = True
+
+
 class ExtractDpkgLogTestCase(unittest.TestCase):
 
     def setUp(self):
         self.tmpdir = tempfile.mkdtemp()
         self.addCleanup(shutil.rmtree, self.tmpdir)
+        self.addCleanup(logging.shutdown)
 
     def test_get_dpkg_log_content(self):
         logfile_dpkg = os.path.join(self.tmpdir, "apt-term.log")
@@ -71,7 +83,8 @@ class CommonTestsForMailxAndSendmail(object):
     def common_setup(self):
         self.tmpdir = tempfile.mkdtemp()
         self.addCleanup(shutil.rmtree, self.tmpdir)
-
+        self.addCleanup(logging.shutdown)
+        logging.root.handlers = []
         # monkey patch to make it testable
         unattended_upgrade.REBOOT_REQUIRED_FILE = os.path.join(
             self.tmpdir, "reboot-required")
@@ -80,6 +93,7 @@ class CommonTestsForMailxAndSendmail(object):
         # setup mail
         apt_pkg.config.set("Unattended-Upgrade::Mail", "root")
         apt_pkg.config.set("Unattended-Upgrade::MailOnlyOnError", "false")
+        apt_pkg.config.set("Unattended-Upgrade::LogDir", self.tmpdir)
 
     def _return_mock_data(self, successful=True):
         """ return input tuple for send_summary_mail """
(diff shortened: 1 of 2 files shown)


test/test_mail.py
```diff
@@ -209,6 +225,24 @@ def test_summary_mail_from_address(self):
...
+        exception_string = "Test exception for email"
+        mock_main_inner.side_effect = Exception(exception_string)
+        # run it
+        options = MockOptions()
+        unattended_upgrade.LOCK_FILE = "./u-u.lock"
+        exception_raised = False
+        try:
+            unattended_upgrade.main(options)
+        except Exception as e:
+            self.assertTrue(str(e) == exception_string)
```
mvo5: You could self.assertEqual() here (not that it matters much). Or even `with self.assertRaises(my_exception):\n...run_thing`.

unattended-upgrade
```diff
@@ -360,6 +360,30 @@ class Unlocked:
             pass
 
 
+class UnattendedUpgradesResult:
```
mvo5: Nice

unattended-upgrade
```diff
@@ -801,7 +825,7 @@ def conffile_prompt(destFile, prefix=""):
         print(_("error message: %s") % e)
         logging.error(_("Apt returned an error, exiting"))
         logging.error(_("error message: %s"), e)
-        sys.exit(1)
+        raise
```
mvo5: Thanks for fixing this!

unattended-upgrade
```diff
@@ -978,9 +1002,16 @@ def _send_mail_using_sendmail(from_address, to_address, subject, body):
     return ret
 
 
-def send_summary_mail(pkgs, res, pkgs_kept_back, pkgs_removed,
-                      pkgs_kept_installed, mem_log, dpkg_log_content):
-    # type: (str, bool, List[str], List[str], List[str], StringIO, str) -> None
+def send_summary_mail(pkgs,                 # type: List[str]
```
mvo5: Probably not for this PR but I wonder if send_summary_mail should simply take the new UnattendedUpgradeResult class?

unattended-upgrade
```diff
@@ -1487,12 +1519,70 @@ def main(options, rootdir=""):
...
+                         install_start_time)
+        if res.result_str and not options.dry_run:
+            # there is some meaningful result which is worth an email
+            log_content = get_dpkg_log_content(logfile_dpkg,
+                                               install_start_time)
+            send_summary_mail(res.pkgs, res.success, res.result_str,
+                              res.pkgs_kept_back, res.pkgs_removed,
+                              res.pkgs_kept_installed, mem_log,
+                              log_content)
+        if (res.update_stamp):
```
mvo5: These `()` are not needed.

unattended-upgrade
```diff
@@ -1487,12 +1519,70 @@ def main(options, rootdir=""):
...
+            if not options.dry_run:
+                # check if the user wants a reboot
+                reboot_if_requested_and_needed()
+        os.close(shutdown_lock)
+        if res.success:
+            return 0
+        else:
+            return 1
+
+    except Exception as e:
```
mvo5: You could probably use  `sys.excepthook` instead of splitting main/main_innter but its fine this way.

unattended-upgrade
```diff
@@ -1487,12 +1519,70 @@ def main(options, rootdir=""):
...
+                                           install_start_time)
+        if not options.dry_run:
+            send_summary_mail(["<unknown>"], False, _("An error occurred"), [],
+                              [], [],
+                              mem_log, log_content)
+        # Re-raise exceptions for apport
+        raise
+
+
+def main_inner(options,             # type: Options
```
mvo5: I would love a better name here,.

test/test_mail.py
rbalint: @mvo5  I tried assertRaises first, but it catches the first exception raised in the code even when it is handled and the code raises a different one finally.
AssertEqual is nicer indeed, I'll fix that.

unattended-upgrade
rbalint: @mvo5 I had the same thinking but send_summary_mail() does not consume update_stamp thus I just left it this way. Also that should be a separate commit which can come later.

unattended-upgrade
rbalint: How about run()?

PR #134, opened 2018-07-04
Speed up u-u by reworking candidate adjustment and precompiling regexps

rbalint:


unattended-upgrade
```diff
@@ -610,9 +598,11 @@ def upgrade_in_minimal_steps(cache,            # type: apt.Cache
             return False
         pkg = cache[pkgname]
         if pkg.is_upgradable:
-            pkg.mark_upgrade(from_user=not pkg.is_auto_installed)
+            cache.call_adjusted(apt.package.Package.mark_upgrade, pkg,
```
mvo5: I wonder if it would be cleaner API wise to add a helper to UnattendedUpgradesCache like "cache.mark_upgrade(pkg)" and have "call_adjusted" just as an internal implementation detail in the cache? This would probably make the code in the bits outside the cache easier to read.

unattended-upgrade
```diff
@@ -1681,8 +1702,11 @@ def main(options, rootdir=""):
         # redo the selection about the packages to upgrade based on the new
         # blacklist
         logging.debug("blacklist: %s" % blacklisted_pkgs)
+        blacklist = [re.compile(pattern) for pattern in blacklisted_pkgs]
```
mvo5: This is nice! I wonder how much speed up this brings in itself.

rbalint: Well, actually nothing in practice, I just re-run the tests on AC power several times and the original measured speed-up was observed due to the low number of samples. Dropping this commit as it is not worth the added complexity.

rbalint: In the re-run the full speedup was 45s -> 4.4s with the commits up to the candidate adjustment changes.

unattended-upgrade
rbalint: Done.

PR #135, opened 2018-07-05
Recommend overriding configuration in a separate file

rbalint:
This can be better than changing /etc/apt/apt.conf.d/50unattended-upgrades
because package updates don't conflict with local changes this way.


rbalint:
@mvo5 added that, too

PR #136, opened 2018-07-06
Further speed improvements and releasing 1.4

rbalint:
Now u-u finishes under 3s when no package needs to be upgraded even when there are packages to be kept back in not allowed origins.
To have another 2x speedup python-apt need to be changed IMO:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=903108


rbalint:
With caching autoremovable packages run time is down to 1.65s and @julian-klode has ideas on using on using lower-level api of python-apt and for speeding up python-apt itself rather then introducing new API.

PR #137, opened 2018-07-10
Fixing autopkgtests for Debian unstable and fix regression

rbalint:


rbalint:
autopkgtest is also passing

debian/unattended-upgrades.service
```diff
@@ -1,14 +1,16 @@
 [Unit]
 Description=Unattended Upgrades Shutdown
+DefaultDependencies=no
+Before=shutdown.target
```
mvo5: Nice

unattended-upgrade
```diff
@@ -157,7 +157,7 @@ class UnattendedUpgradesCache(apt.Cache):
         """
         try:
             new_cand = ver_in_allowed_origin(pkg, self.allowed_origins)
-            if new_cand != pkg.candidate:
+            if new_cand.version < pkg.candidate.version:
```
mvo5: A comment here why "<" vs "!=" would be good I think. I understand this helps with the case when people whitelist experimental if thats true a comment would be good still.

PR #139, opened 2018-08-28
Reopen Cache after commit() even when frontend locking is supported and minor changes

rbalint:
This fixes build and operation with latest python-apt.

### DIFF
--- a/unattended-upgrade
+++ b/unattended-upgrade
@@ -189,6 +189,8 @@ class UnattendedUpgradesCache(apt.Cache):
         function(pkg, **kwargs)
         changes = self.get_changes()
         for marked_pkg in changes:
+            if marked_pkg.name in self._cached_candidate_pkgnames:
+                continue
             if not is_allowed_origin(marked_pkg.candidate,
                                      self.allowed_origins):
                 try:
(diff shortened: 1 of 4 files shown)


unattended-upgrade
```diff
@@ -571,7 +573,7 @@ def cache_commit(cache,           # type: apt.Cache
         else:
             with Unlocked():
                 res = cache.commit(install_progress=iprogress)
-                cache.open()
+        cache.open()
```
mvo5: Nice catch!

unattended-upgrade
```diff
@@ -986,19 +988,6 @@ def wrap(t, width=70, subsequent_indent=""):
     return out
 
 
-def setup_apt_listchanges(conf="/etc/apt/listchanges.conf"):
```
mvo5: Why is this dropped? Is apt-listchanges behaving differently these days? This was (a long time ago :) fixing a real bug, see debian 579733. In the old days APT_LISTCHANGES_FRONTEND=none was needed because if people had set their frontend to pager iirc apt-listchanges would block an unattended-upgrade run because it would present the changes in a pager waiting for someone to confirm them. Is this no longer true?

rbalint: The patch was offered in https://bugs.launchpad.net/ubuntu/+source/unattended-upgrades/+bug/1633098 , but since the bugs were about sending mails and not about blocking I did not check the potential regression with pager. For now I'm dropping this change from the PR and if pager works well, too, I'm scheduling it for a later release.

PR #144, opened 2018-10-03
Fix crashes reported to errors.ubuntu.com and workaround for debootstrap failing

rbalint:


debian/control
```diff
@@ -12,7 +12,7 @@ Build-Depends: debhelper (>= 9.20160709),
                python3-setuptools
 Build-Depends-Indep: python3-dev,
                      python3-coverage,
-                     pep8,
+                     pycodestyle,
```
mvo5: pycodestyle looks like a nice enhancement. I see two potential problems: 
1. Its only available in bionic and later so backporting u-u becomes harder.
2. Its in universe but I guess that is a problem you will deal with :)

unattended-upgrade
```diff
@@ -1403,23 +1403,28 @@ def get_dpkg_log_content(logfile_dpkg, install_start_time):
         logfile_dpkg, install_start_time))
     content = []
     found_start = False
-    with io.open(logfile_dpkg, encoding='utf-8', errors='replace') as fp:
```
mvo5: Nice catch

unattended-upgrade-shutdown
```diff
@@ -89,7 +89,11 @@ def signal_stop_unattended_upgrade():
     if os.path.exists(pidfile):
         pid = int(open(pidfile).read())
         logging.debug("found running unattended-upgrades pid %s" % pid)
-        os.kill(pid, signal.SIGTERM)
+        try:
```
mvo5: Nice one!

debian/control
rbalint: @mvo5 since Travis was also failing I added a fallback, this will make backporting easier.

PR #147, opened 2018-10-09
Fix crash with mirror:// sources and loosen dependencies

rbalint:


unattended-upgrade
```diff
@@ -817,6 +844,15 @@ def check_changes_for_sanity(cache, allowed_origins, blacklist, whitelist,
                 return False
             if not is_pkg_change_allowed(pkg, blacklist, whitelist):
                 return False
+            if (pkg.marked_install
+                and "dkms" in transitive_dependencies(pkg, cache)
```
rbalint: @vorlonofportland this is a bit different workaround than suggested in https://bugs.launchpad.net/ubuntu/+source/shim-signed/+bug/1726803

unattended-upgrade
```diff
@@ -806,6 +832,14 @@ def check_changes_for_sanity(cache, allowed_origins, blacklist, whitelist,
     return True
 
 
+def is_deb(file):
+    # type: (str) -> bool
+    if re.match("^.*\\.deb$", file):
```
mvo5: You could use `file.endswith(".deb")` here, no?

unattended-upgrade
```diff
@@ -422,6 +428,14 @@ def signal_handler(signal, frame):
     SIGNAL_STOP_REQUEST = True
 
 
+def log_once(msg):
```
mvo5: This is fine - I personally would probably do it slightly different, I think its slightly nicer to have a `on_ac_power()` function and then on startup this can be dynamically patched. Something like:
```
import subprocess
import shutil

if not shutil.which("on_ac_power"):
    print("Checking if system ...")
    global on_ac_power
    on_ac_power = lambda: True


def on_ac_power():
    return subprocess.call("on_ac_power") == 1


if __name__ == "__main__":
    print(on_ac_power())
```

unattended-upgrade
```diff
@@ -431,15 +445,27 @@ def should_stop():
...
+            _("Checking if system is running on battery is skipped. Please "
+              "install powermgmt-base package to check power status and skip "
+              "installing updates when the system is running on battery."))
     if apt_pkg.config.find_b(
             "Unattended-Upgrade::Skip-Updates-On-Metered-Connections", True):
-        if NetworkMonitor.get_network_metered(NetworkMonitor.get_default()):
-            logging.warning(_("System is on metered connection, stopping"))
-            return True
+        try:
+            if NetworkMonitor.get_network_metered(
```
mvo5: Same here as above, a helper like "is_metered()" and then dynamically patching it. Feels slightly nicer to me because then reading the checks is less distracting (because now they include the error condition) - hope this explanation makes sense(?). But might be personal preferences and certainly not a blocker.

unattended-upgrade
rbalint: Indeed, fixed.

unattended-upgrade
rbalint: I did consider something like that but please note that with the proposed code the warning is not shown if the check is not configured. Factoring out checks from should stop would indeed make the code easier to read, but I'd to that later.

PR #148, opened 2018-10-18
Running unattended-upgrades-shutdown _before_ shutdown and multiple smaller fixes

rbalint:


unattended-upgrade
```diff
@@ -2062,13 +2062,17 @@ if __name__ == "__main__":
     # init the options
     parser = OptionParser()
     parser.add_option("-d", "--debug",
-                      action="store_true", default=False,
+                      action="store_true",
+                      default=apt_pkg.config.find_b(
+                          "Unattended-Upgrade::Debug", False),
```
mvo5: Nice!

andreyv:
This change has the side effect of the `unattended-upgrade-shutdown --wait-for-signal` process running all the time while the system is up.

Is it fine to disable/mask `unattended-upgrades.service` if `Unattended-Upgrade::InstallOnShutdown` is set to false?

rbalint:
@andreyv It is not fine in the sense that this service is responsible for gracefully stopping u-u _before_ shutdown. Please check the referenced bugs for details.

andreyv:
@rbalint Thanks for the explanation. As I see from the code, `unattended-upgrade-shutdown` performs two functions, possibly stopping a running upgrade process and possibly running upgrades on shutdown.

For the first function, though, why doesn't the running upgrade process take the inhibitor lock itself and listen for `PrepareForShutdown()` events? This way no persistent background process is needed and there is no active inhibitor lock when upgrades aren't actually running.

rbalint:
@andreyv making u-u to take the inhibitor lock could be a useful improvement and I opened https://github.com/mvo5/unattended-upgrades/issues/176 to track that.

PR #149, opened 2018-10-25
 Do not run on development releases until 21 days before release + Ubuntu delta

rbalint:


test/test_dev_release.py
```diff
@@ -0,0 +1,52 @@
...
+
+        # run it
+        options = MockOptions()
+        unattended_upgrade.DISTRO_DESC = "Artful Aardvark (development branch)"
+        unattended_upgrade.main(options, rootdir=self.rootdir)
+        # read the log to see what happend
+        with open(self.log) as f:
+            needle = "Not running on the development release."
+            haystack = f.read()
+            self.assertTrue(needle in haystack,
```
mvo5: Thanks for adding this test

unattended-upgrade
```diff
@@ -1655,6 +1696,37 @@ def run(options,             # type: Options
     if should_stop():
         return UnattendedUpgradesResult(False)
 
+    # check to see if want to auto-upgrade the devel release
+    if apt_pkg.config.find("Unattended-Upgrade::DevRelease") == "auto":
+        try:
+            import distro_info
```
mvo5: Having a test for this would be awesome, maybe we can mock distro_info relatively easily?

PR #156, opened 2018-11-19
Stop using ActionGroups, they interfere with apt.Cache.clear()

rbalint:
causing all autoremovable packages to be handled as newly autoremovable ones
and be removed by default. Dropping ActionGroup usage does not slow down the
most frequent case of not having anything to upgrade and when ther are
packages to upgrade the gain is small compared to the actual package
installation.

Also collect autoremovable packages before adjusting candidates because that
also changed .is_auto_removable attribute of some of them.

LP: #1803749
Closes: #910874

### DIFF
--- a/unattended-upgrade
+++ b/unattended-upgrade
@@ -135,7 +135,6 @@ class UnattendedUpgradesCache(apt.Cache):
     def __init__(self, rootdir, allowed_origins):
         # type: (str, List[str]) -> None
         self._cached_candidate_pkgnames = set()  # type: AbstractSet[str]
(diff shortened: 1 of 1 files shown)


julian-klode:
> Also collect autoremovable packages before adjusting candidates because that
> also changed .is_auto_removable attribute of some of them.

is that safe? I mean, after you adjusted candidates you might end up with a different set of packages that are now autoremovable, because the new candidate depends on other packages.

julian-klode:
Otherwise, it looks good to me.

PR #157, opened 2018-11-20
Fix location of shutdown lock with rootdir, and enable test_dev_relea…

julian-klode:
…se.py

This failed because u-u was using the host's lock file, instead of
one inside the temporary root.


julian-klode:
PID_FILE and PROGRESS_LOCK also exist as absolute paths, but seem unused. Could delete them as well

PR #158, opened 2018-11-26
Fix blacklists, clear cache only when needed, add --no-minimal-steps, improve tests and fix regressions

rbalint:
When collecting upgradable packages the upgradable ones stayed in the
cache and they were upgraded together even when unattended-upgrades
was configured to perform upgrades in minimal steps.

### DIFF
--- a/data/50unattended-upgrades.Debian
+++ b/data/50unattended-upgrades.Debian
@@ -41,12 +41,26 @@ Unattended-Upgrade::Origins-Pattern {
 //      "o=Debian Backports,a=${distro_codename}-backports,l=Debian Backports";
 };
 
-// List of packages to not update (regexp are supported)
+// Python regular expressions, matching packages to exclude from upgrading
 Unattended-Upgrade::Package-Blacklist {
-//     "vim";
-//     "libc6";
-//     "libc6-dev";
-//     "libc6-i686";
+    // The following matches all packages starting with linux-
+//  "linux-";
+
+    // Use $ to explicitely define the end of a package name. Without
+    // the $, "libc6" would match all of them.
+//  "libc6$";
+//  "libc6-dev$";
+//  "libc6-i686$";
+
+    // Special characters need escaping
+//  "libstdc\+\+6$";
+
+    // The following matches packages like xen-system-amd64, xen-utils-4.1,
+    // xenstore-utils and libxenstore3.0
+    "(lib)?xen(store)?";
+
+    // For more information about Python regular expressions, see
+    // https://docs.python.org/3/howto/regex.html
 };
 
 // This option allows you to control if on a unclean dpkg exit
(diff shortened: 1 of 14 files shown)


data/50unattended-upgrades.Debian
```diff
@@ -41,12 +41,26 @@ Unattended-Upgrade::Origins-Pattern {
...
+// Python regular expressions, matching packages to exclude from upgrading
 Unattended-Upgrade::Package-Blacklist {
-//     "vim";
-//     "libc6";
-//     "libc6-dev";
-//     "libc6-i686";
+    // The following matches all packages starting with linux-
+//  "linux-";
+
+    // Use $ to explicitely define the end of a package name. Without
```
mvo5: Nice!

kernel/postinst.d/unattended-upgrades
```diff
@@ -6,5 +6,8 @@ case "$DPKG_MAINTSCRIPT_PACKAGE::$DPKG_MAINTSCRIPT_NAME" in
 esac
 
 if [ -d /var/run ]; then
-   touch /var/run/reboot-required
+    touch /var/run/reboot-required
+    if ! grep -q "^$DPKG_MAINTSCRIPT_PACKAGE$" /var/run/reboot-required.pkgs; then
```
mvo5: :+1:

unattended-upgrade
```diff
@@ -140,6 +141,9 @@ class UnattendedUpgradesCache(apt.Cache):
         self.allowed_origins = allowed_origins
         apt.Cache.__init__(self, rootdir=rootdir)
 
+        # pre-heat lazy-loaded modules to avoid crash on python upgrade
```
mvo5: Thanks for the comment!

unattended-upgrade
```diff
@@ -1574,8 +1570,8 @@ def do_auto_remove(cache,             # type: apt.Cache
     else:
         for pkgname in auto_removable:
             cache[pkgname].mark_delete()
-        if is_autoremove_valid(cache, "", auto_removable, blacklisted_pkgs,
-                               whitelisted_pkgs):
+        if is_autoremove_valid(cache, "", auto_removable, blacklist,
```
mvo5: This can probably now be a single line (probably <79 chars now)?

unattended-upgrade
```diff
@@ -1447,8 +1440,12 @@ def calculate_upgradable_pkgs(cache,             # type: apt.Cache
                            pkgs_kept_back,
                            cache,
                            allowed_origins,
-                           blacklisted_pkgs,
-                           whitelisted_pkgs)
+                           blacklist,
+                           whitelist)
+
+    if (cache.get_changes()):
```
mvo5: (nitpick) No need for the extra `()`, i.e. `if cache.get_changes():` is fine :)

unattended-upgrade
rbalint: It would become == 80, but tried that, too :-)

unattended-upgrade
rbalint: Fixed that, thanks!

PR #159, opened 2018-11-27
Test dev release auto

julian-klode:
Depends on/Incorporates #157


test/test_dev_release.py
```diff
@@ -21,19 +21,35 @@ class MockOptions(object):
...
+        unattended_upgrade.LOCK_FILE = "./u-u.lock"
         self.rootdir = os.path.abspath("./root.untrusted")
         self.log = os.path.join(
             self.rootdir, "var", "log", "unattended-upgrades",
             "unattended-upgrades.log")
 
+        self.apt_conf = os.path.join(self.rootdir, "etc", "apt",
+                                     "apt.conf")
+
+        os.rename(self.apt_conf, self.apt_conf + ".bak")
```
rbalint: would not it be nicer to use a fragment in apt.conf.d/ ?

julian-klode: I don't know, I feel like we should do some generic refactoring of the test rootdir interaction.

rbalint: well, i share the feeling then :-)
it is fine for now

rbalint:
@julian-klode Merging was a bit early, autopkgtest fails in unstable and disco when building from source :-(

PR #160, opened 2018-11-29
Fix the positive test case for test-dev-release

julian-klode:
Add a new message stating that we are running on the devel
release, and check for its existence. The message is awfully
similar to the "Not running" messages, though, might want to
improve that a bit


rbalint:
Thanks, I tested it, too.

PR #164, opened 2018-12-30
Auto-retrieve repositories to be updated

tuncK:
I tried to add a new functionality: I would like to generate a tailor-made /etc/apt/apt.conf.d/50unattended-upgrades file during the installation, so that the user automagically has the repos that he/she were already updating by apt-get by default. This might make it more user-friendly.

### DIFF
--- a/debian/preinst
+++ b/debian/preinst
@@ -9,6 +9,13 @@ set -e
 
 CONFIG="/etc/apt/apt.conf.d/50unattended-upgrades"
 
+# Retrieves (mostly user-defined) repositories in /apt/lists 
+# and configures for automatic updates.
+missingSources=$(python listMissingRepos.py)
+for source in $missingSources; do
+    sed -i "/Unattended-Upgrade::Allowed-Origins {/a \\\t$source" $CONFIG
+done
+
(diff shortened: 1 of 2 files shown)


julian-klode:
that's a terrible idea. the whole point of the list is to only install security updates.

debian/preinst
```diff
@@ -9,6 +9,13 @@ set -e
 
 CONFIG="/etc/apt/apt.conf.d/50unattended-upgrades"
 
+# Retrieves (mostly user-defined) repositories in /apt/lists 
+# and configures for automatic updates.
+missingSources=$(python listMissingRepos.py)
+for source in $missingSources; do
+    sed -i "/Unattended-Upgrade::Allowed-Origins {/a \\\t$source" $CONFIG
```
rbalint: Handling configuration files is not that easy. If you would like to contribute to Debian packages in that area please read about dpkg's configuration file handling mechanism and ucf.

rbalint:
I don't think that this feature is useful either. System administrators can take a look at the output of ```apt-cache policy``` to pick a particular origin to enable or add an origin pattern in a separate configuration file that enables every available origin.

PR #173, opened 2019-03-05
Remove double "format" in 50unattended-upgrades files

brannondorsey:
A small grammatical fix in the comments of the default `50unattended-upgrades` files.


rbalint:
Thanks!

PR #175, opened 2019-03-21
Candidate adjustment-related and other minor fixes

rbalint:


unattended-upgrade
```diff
@@ -1119,6 +1182,9 @@ def send_summary_mail(pkgs,                 # type: List[str]
     if (res and apt_pkg.config.find_b(
             "Unattended-Upgrade::MailOnlyOnError", False)):
         return
+    # if the run was successful but nothing had to be done skip sending email
+    if (res and not pkgs and not pkgs_kept_back and not pkgs_removed):
```
mvo5: A test for this would be nice to prevent this from happening again

unattended-upgrade
```diff
@@ -204,9 +228,44 @@ class UnattendedUpgradesCache(apt.Cache):
...
+            # with the current candidate adjustments.
+            # Try marking it without the adjustments then adjust
+            # all packages to be installed/upgraded. This will work where
+            # package dependencies are similar for the adjusted and not
+            # adjusted case, just with different versions.
+            logging.debug("falling back to marking %s, then adjusting changes"
+                          % pkg.name)
+            for pkg2 in ([self[pkg_name] for pkg_name
+                          in self._cached_candidate_pkgnames]):
+                pkg2.candidate = pkg2.versions[0]
```
julian-klode: Um, why is this resetting all candidates to to their 0th version (which is IIRC, always the latest available version)? That seems wrong.

rbalint: @julian-klode Typically the highest versions are installable with the highest versions of the dependencies, thus the package set would be pulled from -updates. Later if that work we try to adjust the set to versions from allowed origins and if that worked we are ok.
The reason of that is sudo apt install systemd=237-3ubuntu10.13 fails because .15 is available from updates and needs strict versioned dependencies.

debian/changelog
```diff
@@ -1,3 +1,21 @@
+unattended-upgrades (1.12) UNRELEASED; urgency=medium
+
+  [ Brannon Dorsey ]
+  * Remove double "format" in 50unattended-upgrades configuration files
+
+  [ Balint Reczey ]
+  * Compare apt.package.Version objects and not the versions' string
```
mvo5: This is a nice catch and it explains the missing libnss-systemd upgrade!

unattended-upgrade
```diff
@@ -178,7 +178,7 @@ class UnattendedUpgradesCache(apt.Cache):
             # Only adjust to lower versions to avoid flipping back and forth
             # and to avoid picking a newer version, not selected by apt.
             # This helps avoiding upgrades to experimental's packages.
-            if new_cand.version < pkg.candidate.version:
+            if new_cand < pkg.candidate:
```
mvo5: Nice catch!  I wish we had a test for this.

debian/changelog
```diff
@@ -1,3 +1,21 @@
...
+    representation. This prevented adjusting candidates when the strings sorted
+    differently (LP: #1820888)
+  * Enable test_clean and test_patch_days tests
+  * Fall back to adjusting more packages' candidates
```
mvo5: I'm not sure about this change, I don't fully understand why the regular adjust_candidate logic fails here. I would like to understand this a bit better as this seems to add quite a bit of complexity. An example where this is needed would be nice. I mean, the pockets should be self-contained, i.e. it should be possible to install a package if only "foo-security" but not "foo-updates" is enabled. So I'm confused where the failure comes from.

unattended-upgrade
rbalint: I added the test to autopkgtests.

debian/changelog
rbalint: The original logic fails, because pkg.mark_upgrade() fails when pkg's candidate is adjusted to -security from -updates and pkg (from -security) versioned depends only on something not present in -updates (with the required version).

Apt command line also fails here on Bionic:
```
sudo apt install systemd=237-3ubuntu10.13
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 systemd : Depends: libsystemd0 (= 237-3ubuntu10.13) but 237-3ubuntu10.15 is to be installed
           Recommends: libpam-systemd but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```

The original adjusting logic assumed this to work.

rbalint: If one package does not depend with (= N) on the others then the set may be set for installation when u-u tries to start with the package with weaker dependencies.

unattended-upgrade
julian-klode: OK

debian/changelog
julian-klode: FWIW, I think the more correct approach would be to mark the change for upgrade, then look at which packages are broken by that change and adjust the candidates of those too match. Or just follow the dependencies and switch all of them to security.

apt does have similar code to handle this, so, e.g apt install systemd/bionic-security is supposed to follow dependencies and switch those to security as well if they don't match otherwise.

rbalint: I agree that building a proper resolver would fix the issue, but I don't agree that it should be implemented in u-u, rather than in apt.
Trying to resolve the situation by walking the dependency chain would fail when the package can't be marked to upgrade do to a Breaks: set in an other package installed/to be installed.
This is a workaround for apt's resolver failing to find a solution for installing systemd/bionic-security . 
IMO the correct approach is making u-u use and respect pinning and fixing apt's resolver, but none of those are ready and the two fallbacks are expected to handle most package expected dependency constellations in the archive.

rbalint: Also thanks for pointing it out that `apt-get install systemd/bionic-security` works because `apt-get install systemd=237-3ubuntu10.13` does not and I only tried the latter.
Can I somehow do the former via python-apt?

rbalint:
Got ack via IRC from @julian-klode and @mvo5, thus merging it myself.

debian/changelog
rbalint: @julian-klode OK, since the full fallback is super slow and it gets triggered on held packages I went with marking dependencies: https://github.com/mvo5/unattended-upgrades/pull/190

See: https://bugs.launchpad.net/ubuntu/+source/unattended-upgrades/+bug/1824804

PR #177, opened 2019-03-25
Improve the email report and minor fixes

rbalint:
The kept packages would look like that in the email:
...
Packages with upgradable origin but kept back:
 Ubuntu cosmic-security:
  libavahi-client3
 LP-PPA-rbalint-scratch:
  other-package
...

### DIFF
--- a/unattended-upgrade
+++ b/unattended-upgrade
@@ -73,6 +73,8 @@ from subprocess import (
     Popen,
     PIPE,
 )
+from textwrap import wrap
+
 import apt
 import apt_inst
 import apt_pkg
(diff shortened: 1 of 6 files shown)


unattended-upgrade
```diff
@@ -453,6 +454,38 @@ class Unlocked:
...
+    """
+    def add(self, pkg, cache):
+        # type: (apt.Package, UnattendedUpgradesCache) -> None
+        allowed_origins = cache.allowed_origins
+        for origin in ver_in_allowed_origin(pkg, allowed_origins).origins:
+            if is_allowed_origin(origin, allowed_origins):
+                self[origin.origin + " " + origin.archive].add(pkg.name)
+                return
+
+    def pop_upgradable(self, cache):
```
julian-klode: This is confusing, why would I want to pop packages that are upgradable from a dict origin -> package? Maybe the class needs a better name that explains the purpose better

unattended-upgrade
```diff
@@ -453,6 +456,38 @@ class Unlocked:
...
```
julian-klode: This is confusing, why is that needed? 

Answer: Because we only want to list kept-back packages; but this is not explicit in the name of the class (or method?)

rbalint: How about naming the class KeptPackages and adding some comments? Do you have a better idea for the name?

rbalint: @julian-klode I tried to clear the confusion, do you think it is good now?

julian-klode: Yeah

PR #178, opened 2019-03-26
Fix missing space

AlbanVidal:
Signed-off-by: Alban VIDAL <alban.vidal@zordhak.fr>


julian-klode:
That file is autogenerated from the source code, so the source code needs to be fixed, not the file.

rbalint:
@AlbanVidal Thanks, I included the fix in https://github.com/mvo5/unattended-upgrades/pull/177

PR #185, opened 2019-04-11
Conffile moves

rbalint:
Detect changes to moved conffiles and also not crash


test/test_conffile.py
```diff
@@ -24,13 +24,33 @@ def setUp(self):
 No really.
 """)
 
+    def tearDown(self):
+        try:
```
mvo5: Could you use shutil.rmtree() here? That may simplify the code.

unattended-upgrade
```diff
@@ -1098,14 +1118,21 @@ def conffile_prompt(destFile, prefix=""):
...
                 logging.debug("found conffile %s in new pkg but on dpkg "
                               "status" % conf_file)
                 pkg_md5sum = get_md5sum_for_file_in_deb(destFile, conf_file)
-                with open(prefix + conf_file, 'rb') as fp:
-                    if pkg_md5sum != apt_pkg.md5sum(fp):
-                        return True
+                if os.path.isdir(installed_conf_file):
+                    logging.debug("found conffile %s being a directory on the "
+                                  "system " % conf_file)
+                    pass
```
mvo5: I don't think this `pass` here is needed.

test/test_conffile.py
rbalint: Thanks, fixed that.

unattended-upgrade
rbalint: Good point, I felt that pass here helps following the logic, but it is indeed obsolete.

PR #187, opened 2019-04-12
 Detect unchanged moved conffiles

rbalint:


unattended-upgrade
```diff
@@ -995,24 +998,44 @@ def get_md5sum_for_file_in_deb(deb_file, conf_file):
...
+    if os.path.isdir(prefix + conf_file):
+        logging.debug("found conffile %s is a directory on the system "
+                      % conf_file)
+        return "dir"
+
+    # test against the installed file, if the local file got deleted
+    # by the admin thats ok but it may still trigger a conffile prompt
+    # (see debian #788049)
+    if os.path.exists(prefix + conf_file):
+        with open(prefix + conf_file, 'rb') as fb:
```
julian-klode: Better to put the `open` (well, the entire with block) in a try and then catch either `FileNotFoundError` or `OsError`, rather than using `os.path.exists` IMO.

unattended-upgrade
```diff
@@ -1117,21 +1131,15 @@ def conffile_prompt(destFile, prefix=""):
...
     # in the previous version in the dpkg status file
     if pkg_conffiles:
-        for conf_file in pkg_conffiles.split("\n"):
-            installed_conf_file = find_old_installed_conf_file(
-                conf_file, prefix)
-            if conf_file not in dpkg_status_conffiles \
-               and os.path.exists(installed_conf_file):
+        for conf_file in pkg_conffiles:
+            old_conf_file = map_conf_file(conf_file, dpkg_status_conffiles)
+            if not old_conf_file and os.path.exists(prefix + conf_file):
```
julian-klode: Could we drop the `os.path.exists()` here, as the function `get_md5sum_for_file_installed` already handles that by returning "". Maybe adding `and current_md5 != ""` in the `if` below to make sure we don't get odd cases.

unattended-upgrade
rbalint: Good point, fixing that.

unattended-upgrade
rbalint: Fixed that, too.

julian-klode:
those Changes LGTM

PR #190, opened 2019-04-15
Adjust only transitive dependencies in the fallback

rbalint:
when a package from an allowed origin can't be marked to install/upgrade.

This is a much lighter approach than marking every upgradable package because
the full fallback was triggered on packages held back as well, using an
excessive amount of CPU time.

LP: #1824804

### DIFF
--- a/unattended-upgrade
+++ b/unattended-upgrade
@@ -190,13 +190,6 @@ class UnattendedUpgradesCache(apt.Cache):
         except NoAllowedOriginError:
             return False
 
-    def adjust_candidate_for_all_packages(self):
-        """ Adjust origin of all the upgradable/installable packages
-        """
-        for pkg in self:
-            if pkg.is_upgradable or not pkg.is_installed:
-                self.adjust_candidate(pkg)
-
     def call_checked(self, function, pkg, **kwargs):
         """ Call function and check if package is in the wanted state
         """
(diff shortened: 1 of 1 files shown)


julian-klode:
Wondering if you also want to adjust for transitive Conflicts/Breaks, so that if you have Conflicts: foo (<< 2) and foo (>= 2) available, it also adjusts foo.

rbalint:
Yes, I was wondering about too, but skipped that in the first iteration. How about just not restricting the relationships when collecting deps for adjustment? Should not make a huge difference.

unattended-upgrade
```diff
@@ -863,20 +863,22 @@ def is_pkg_change_allowed(pkg, blacklist, whitelist):
...
 
 
-def transitive_dependencies(pkg, cache, acc=set()):
-    # type (apt.Package, apt.Cache, AbstractSet[str]) -> bool
+def transitive_dependencies(pkg, cache, acc=set(), valid_types=None):
+    # type (apt.Package, apt.Cache, AbstractSet[str], AbstractSet[str]) -> bool
     """ All (transitive) dependencies of the package
     """
     # Note that alternative (|) dependencies are collected, too
     valid_types = {'Depends', 'PreDepends', 'Recommends'}
```
julian-klode: AFAICT; This is still overriding the parameter.

rbalint: Thanks, fixed that!

rbalint:
Indeed Suggests and Enhances can be excluded, but the difference would not worth the extra code IMO.

PR #191, opened 2019-04-15
Stop raising NoAllowedOriginError when marking packages to upgrade/in…

rbalint:
…stall fails

LP: #1824876


rbalint:
I aimed at better error propagation when I added the raise but handling the exception in all callers just became too painful and u-u already checks if the marked changes are OK making the exception redundant.

PR #192, opened 2019-04-15
Skip trying to upgrade held packages in call_adjusted()

rbalint:
LP: #182480


julian-klode:
Just noticed the bug number seems wrong though, gotta fix that one in the changelog.

rbalint:
@mvo5 test_rewind.py still covers the fallback, this is why I have not added new tests and I also hope progressing on moving to pins from adjustments thus the tests would become obsolete.

rbalint:
@julian-klode Thanks, fixed it in the changelog in master

PR #195, opened 2019-04-24
Use mark_install_adjusted() in rewind_cache()

rbalint:
The original cache had packages marked with adjustments thus rewinding
should also do adjustments to reach the same state.

Also not using mark_install_adjusted() crashes when apt raises error on
held packages.

LP: #1826157


rbalint:
I could update the rewind test that it uses adjustments but i would like to drop rewinding completely instead soon.

rbalint:
I pushed the updated test to help SRUs.

PR #199, opened 2019-05-08
Expand kernel patterns

rbalint:


test/autopkgtest_kernel_patterns.py
```diff
@@ -0,0 +1,45 @@
...
+import logging
+import re
+import subprocess
+import unittest
+
+from unattended_upgrade import (running_kernel_pkgs_regexp,
+                                versioned_kernel_pkgs_regexp)
+
+
+class TestSubstitude(unittest.TestCase):
```
julian-klode: "Substitude" -> "Substitut"e

rbalint: Fixed.

rbalint:
@julian-klode I think it may even be better to push the regexp generation logic to pyhon-apt because update-manager needs to do the same, maybe u-r-u, too, to not remove running kernel.

julian-klode:
I agree!

rbalint:
@julian-klode  If LP: #1615381 could be  fixed in libapt first to protect the running kernel then only an API to list the versioned kernel-related autoremovable packages would be needed in python-apt. Then API clients could save the filtering to protect the running kernel.

rbalint:
@julian-klode How about merging this for now to u-u, then dropping it when python-apt exposes API and APT is fixed to do parts of this?

PR #201, opened 2019-05-27
Fix Debian's autopkgtest

rbalint:


debian/tests/upgrade-all-security
```diff
@@ -53,6 +54,12 @@ chroot_exec "$chroot_dir" bash -c 'apt install -y /tmp/unattended-upgrades_*deb'
 
 
 enable_pocket "$chroot_dir" security
+chroot_exec "$chroot_dir" apt-get update
+if ! chroot_exec "$chroot_dir" apt list --upgradable 2> /dev/null | grep -q .; then
```
mvo5: Silly question - the `grep -q .` is to check if there is any output in apt list --upgradable? If so would it maybe make sense to use something like "wc" to make this more explicit? Also when I run this here with an empty upgrade set I get: 
```
LANGUAGE=en apt list --upgradable 2>/dev/null |grep  .  && echo grep-successful
Listing...
grep-successful
```
is that expected?

julian-klode: This does not work, apt list --upgradable include progress output in stdout.

julian-klode: Can you use apt dist-upgrade -s | grep -q ^Inst instead?

rbalint: @mvo5 @julian-klode Thanks, fixing it.

rbalint: Fixed.

PR #203, opened 2019-06-04
Replace the boolean Unattended-Upgrade::MailOnlyOnError mailing repor…

original-birdman:
…t option with a (multi-value) Unattended-Upgrade::MailReport setting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,10 +204,14 @@ Supported Options Reference
  Unattended-Upgrade::Sender "server@example.com";
  ```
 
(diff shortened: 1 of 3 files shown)


original-birdman:
I've changed `test/aptroot/etc/apt/apt.conf.d/50unattended-upgrades` to reflect the new configuration option, but haven't changed any of the `50unattended-upgrades.*` files under `data/`, as I'm unsure who needs to do that.

PR #205, opened 2019-06-04
Replace boolean Unattended-Upgrade::MailOnlyOnError mailing report op…

original-birdman:
…tion with multi-valued (string) Unattended-Upgrade::MailReport.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,10 +204,14 @@ Supported Options Reference
  Unattended-Upgrade::Sender "server@example.com";
  ```
 
-* `Unattended-Upgrade::MailOnlyOnError` - boolean (default:False)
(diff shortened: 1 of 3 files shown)


original-birdman:
I've updated `test/aptroot/etc/apt/apt.conf.d/50unattended-upgrades` with the new option, but haven't changed any of the `50unattended-upgrades.*` files in `data/` as I'm not sure who "owns" them.

rbalint:
@original-birdman Files in `data/` are maintained here for the benefit of the derivative distros. Please update the configuration files, too.

original-birdman:
Re-doing with `data/` files modified as well.

rbalint:
You can just force-push to your branch to get the existing PRs updated, there is no need to open new ones.

PR #214, opened 2019-07-12
Use Debian's security updates from $distro-security starting with Bullseye

rbalint:
See: https://lists.debian.org/debian-devel-announce/2019/07/msg00004.html

Gbp-Dch: Short


julian-klode:
Also needs updating the Debian config from
```
        "origin=Debian,codename=${distro_codename},label=Debian-Security";
```

to
```
        "origin=Debian,codename=${distro_codename}-security,label=Debian-Security";
```

rbalint:
@julian-klode thanks, fixed

PR #227, opened 2019-10-08
* Allow package downgrade if Pin-Priority exceeds 1000

h2x0r:
This PR allows unattended-upgrade to downgrade a package if a Pin-Priority exceeds 1000 and downgrade is allowed (Unattended-Upgrade::Allow-downgrade "true";).  It can be handy on many circumstances including:
1. Unattended rollback:
  - You have your own internal/private repository and thousands of hosts tracking changes on it. In case you want to roll back to previous version it is enough if you just remove latest version from repository
2. Software testing:
  - You can quickly roll back and forward only by adding or removing packages from repository


rbalint:
@h2x0r This is a very interesting feature and aligns with the goal of fully supporting pinning. Please add test cases to let us keep the feature working easily.

debian/changelog
```diff
@@ -1,3 +1,9 @@
+unattended-upgrades (1.14+nmu1) unstable; urgency=low
+
+  * Allow package downgrade if Pin-Priority exceeds 1000
+
+ -- Kaupo Arulo <kaupoa@microsoft.com>  Tue, 08 Oct 2019 13:00:22 +0300
+
```
rbalint: Please drop the change to the changelog, also please don't NMU it.

unattended-upgrade
```diff
@@ -220,8 +220,9 @@ class UnattendedUpgradesCache(apt.Cache):
             self.adjust_candidate(self[pkg_name])
 
         if function == apt.package.Package.mark_upgrade \
-            and not pkg.is_upgradable:
-            if not apt_pkg.config.find_b("Unattended-Upgrade::Allow-downgrade", False):
+                and not pkg.is_upgradable:
```
rbalint: Please squash this commit to the previous one to keep tests happy in every commit.

debian/changelog
h2x0r: Done

unattended-upgrade
h2x0r: Done. Sorry for a long delay...

PR #236, opened 2019-12-13
Keep mypy happy

rbalint:


rbalint:
@julian-klode I think this will be enough for the new python-apt.

Issue #238, opened 2019-12-29
unattended-upgrade crashes after commit 6552cc8 in calculate_upgradable_pkgs


Prototyped:
Hello,

It looks like after commit 6552cc8, on my Debian bullseye desktop computer, unattended-upgrade fails with every run in calculate_upgradable_pkgs(). It looks like the change to add the allow-downgrades toggle has made an assumption that does not hold true on my system.

```shell
[14:44|~/unattended-upgrades]
amitg@athena. git checkout 4163ec7
HEAD is now at 4163ec7 Save a Cache.clear() in download-only mode
[14:44|~/unattended-upgrades]
amitg@athena. sudo ./unattended-upgrade -d --dry-run > unattended-upgrade.log.4163ec7
/usr/bin/dpkg --status-fd 10 --no-triggers --unpack --auto-deconfigure --recursive /tmp/apt-dpkg-install-Y0fgrK
/usr/bin/dpkg --status-fd 10 --configure --pending
/usr/bin/dpkg --status-fd 10 --no-triggers --unpack --auto-deconfigure /var/cache/apt/archives/python3-cffi-backend_1.13.2-1_amd64.deb
/usr/bin/dpkg --status-fd 10 --configure --pending
/usr/bin/dpkg --status-fd 10 --no-triggers --unpack --auto-deconfigure /var/cache/apt/archives/software-properties-common_0.96.20.2-2.1_all.deb /var/cache/apt/archives/python3-software-properties_0.96.20.2-2.1_all.deb
/usr/bin/dpkg --status-fd 10 --configure --pending
# [... many other dpkg command lines ...]
/usr/bin/dpkg --status-fd 10 --no-triggers --unpack --auto-deconfigure /var/cache/apt/archives/syslinux_3%3a6.04~git20190206.bf6db5b4+dfsg1-2_amd64.deb
/usr/bin/dpkg --status-fd 10 --configure --pending
/usr/bin/dpkg --status-fd 10 --no-triggers --unpack --auto-deconfigure /var/cache/apt/archives/libglusterfs0_7.1-1_amd64.deb /var/cache/apt/archives/libgfxdr0_7.1-1_amd64.deb /var/cache/apt/archives/libgfrpc0_7.1-1_amd64.deb /var/cache/apt/archives/libgfapi0_7.1-1_amd64.deb
/usr/bin/dpkg --status-fd 10 --configure --pending
[14:47|~/unattended-upgrades]
[...]

PR #239, opened 2019-12-29
Check that a candidate is available before checking the version has changed.

Prototyped:
# What this change is

This change fixes https://github.com/mvo5/unattended-upgrades/issues/238 where a package installed using dpkg that does not have an upgrade candidate owing to an apt preference, causes calculation of the upgrade to raise an AttributeError. The change involves checking that a candidate exists before checking that the version is different from the installed version.

# Testing done

```shell
sudo ./unattended-upgrade -d --dry-run
```

with a package, alpine-doc, built from alpine source and installed using dpkg, while the following apt preference is set:

```
Package: alpine-doc
Pin: origin www.mirrorservice.org
Pin-Priority: -10
```

The command invocation now succeeds where previously unattended-upgrades crashed with an `AttributeError: 'NoneType' object has no attribute 'version'`.

Also, tests are passing, invoked using

```shell
make -C test
```

### DIFF
--- a/unattended-upgrade
+++ b/unattended-upgrade
@@ -255,7 +255,7 @@ class UnattendedUpgradesCache(apt.Cache):
                     ver_in_allowed_origin(marked_pkg,
                                           self.allowed_origins)
                     # important! this avoids downgrades below
-                    if pkg.is_installed and not pkg.is_upgradable and\
+                    if pkg.is_installed and not pkg.is_upgradable and \
                             apt_pkg.config.find_b("Unattended-Upgrade::Allow-"
                                                   "downgrade", False):
                         continue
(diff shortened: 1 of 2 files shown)


unattended-upgrade
```diff
@@ -1611,6 +1608,14 @@ def try_to_upgrade(pkg,               # type: apt.Package
         pkgs_kept_back.add(pkg, cache)
 
 
+def has_candidate_pkg_version_changed(pkg,       # type: apt.Package
```
rbalint: I think `candidate_version_changed` would be a still descriptive but simpler name.

Prototyped: Function renamed.

Prototyped:
Function renamed as requested.

PR #250, opened 2020-02-01
Fix syntax in template conf files

louib:


data/50unattended-upgrades.md5sum
```diff
@@ -103,3 +103,6 @@ cdbd918a6df20b697f0837fb68ead9dd
 b3c34754524bf71ea49ed488f956e0d5
 a599aee6e3f9dcadf2189683accc6c9e
 6cfaf67893454c604aaef30a57e63521
+4f04cda01dec68446f7b1355c78a8b97
+9daa6698ec59407b38f30c7a65fe4dc9
+c6629e6c7b79dff9d34d200c42afe757
```
rbalint: Thanks for taking care of the md5sums, too, but please skip this commit from the PR.
It is enough to update the md5sums once per new release and there may be other changes to the config files

louib: @rbalint the CI build was failing without those new checksums, that's why I added them. I you don't care I can remove them anyway.

rbalint: @louib Yes, but if you rebase your branch to current master CI won't fail.

Issue #223, opened 2019-09-18
sanity check failed for ...


albertnetymk:
On logging into my system, I was greeted by the following:

```
2 updates could not be installed automatically. For more details,
see /var/log/unattended-upgrades/unattended-upgrades.log
```
The log doesn't seem to contain anything insightful, so I decided to debug it using `-d`. Unfortunately, it's not obvious to me why the sanity check fails. Anyone can give me some instruction on how I can proceed to debug this?
```
$ unattended-upgrade -d -v --dry-run`
Checking if system is running on battery is skipped. Please install powermgmt-base package to check power status and skip installing updates when the system is running on battery
.
Initial blacklist :
Initial whitelist:
Starting unattended upgrades script
Allowed origins are: o=Debian,a=testing, o=Debian,a=testing-updates
Using (^linux-image-[1-9][0-9]*\.[0-9]+\.[0-9]+-[0-9]+(-.+)?$|^linux-headers-[1-9][0-9]*\.[0-9]+\.[0-9]+-[0-9]+(-.+)?$|^linux-image-extra-[1-9][0-9]*\.[0-9]+\.[0-9]+-[0-9]+(-.+)?
$|^linux-modules-[1-9][0-9]*\.[0-9]+\.[0-9]+-[0-9]+(-.+)?$|^linux-modules-extra-[1-9][0-9]*\.[0-9]+\.[0-9]+-[0-9]+(-.+)?$|^linux-signed-image-[1-9][0-9]*\.[0-9]+\.[0-9]+-[0-9]+(-
.+)?$|^linux-image-unsigned-[1-9][0-9]*\.[0-9]+\.[0-9]+-[0-9]+(-.+)?$|^kfreebsd-image-[1-9][0-9]*\.[0-9]+\.[0-9]+-[0-9]+(-.+)?$|^kfreebsd-headers-[1-9][0-9]*\.[0-9]+\.[0-9]+-[0-9
]+(-.+)?$|^gnumach-image-[1-9][0-9]*\.[0-9]+\.[0-9]+-[0-9]+(-.+)?$|^.*-modules-[1-9][0-9]*\.[0-9]+\.[0-9]+-[0-9]+(-.+)?$|^.*-kernel-[1-9][0-9]*\.[0-9]+\.[0-9]+-[0-9]+(-.+)?$|^lin
[...]

PR #255, opened 2020-02-22
Collect kept packages later

rbalint:


rbalint:
@julian-klode this will be needed when the applied pinning will hide why a packages is not marked to be installed

rbalint:
This fixes #223 .

PR #260, opened 2020-03-11
Keep adjusting packages

rbalint:


rbalint:
@julian-klode this seems to be a good compromise on speed and still fixes resolver issues in the fallback

PR #272, opened 2020-05-09
Speed improvements

rbalint:


mvo5:
Thanks! This looks fine, out of curiosity, how much speed improvements does this bring?

rbalint:
> Thanks! This looks fine, out of curiosity, how much speed improvements does this bring?

It depends. The regex fix is barely noticeable in real life, but the issue fixed in https://github.com/mvo5/unattended-upgrades/pull/272/commits/9dfc7dcda1b79aa71a567eeac2648a298f2e9e29 can hold up u-u for hours when many packages are held back and many packages could be upgraded.

PR #277, opened 2020-06-16
Fix strict whitelist

rbalint:


rbalint:
A test suite extension would be nice, maybe I do it before merging.

PR #280, opened 2020-07-21
Make APT marking fallback configurable and disable it on Sid

rbalint:
Closes: #960966


rbalint:
@julian-klode How do you like this workaround?

Issue #295, opened 2021-05-03
unattended-upgrade-shutdown --wait-for-signal doesn't work


kucharskim:
A while ago I configured multiple Ubuntu 20.04.2 LTS machines with unattended upgrade on shutdown:

```
# grep -ve '^[[:blank:]]*//' -e '^[[:blank:]]*$' /etc/apt/apt.conf.d/75upgrades-custom 
Unattended-Upgrade::Origins-Pattern {
        "site=*.ubuntu.com,n=${distro_codename}";
        "site=packages.microsoft.com,n=stable";
        "site=dl.google.com,n=stable";
};
Unattended-Upgrade::MinimalSteps "true";
Unattended-Upgrade::Remove-Unused-Dependencies "true";
Unattended-Upgrade::Remove-Unused-Kernel-Packages "true";
Unattended-Upgrade::InstallOnShutdown "true";
```

After few weeks I've noticed that all of them have pending upgrades not installed. I spent some time and I think problem is with following line:

```
diff --git unattended-upgrade-shutdown unattended-upgrade-shutdown
index f59e212..8c30a53 100755
--- unattended-upgrade-shutdown
+++ unattended-upgrade-shutdown
@@ -339,7 +339,7 @@ class UnattendedUpgradesShutdown():
         res = apt_pkg.get_lock(self.options.lock_file)
         logging.debug("get_lock returned %i" % res)
         # exit here if there is no lock
-        if res > 0:
+        if res < 0:
             logging.debug("lock not taken")
             if self.lock_was_taken:
                 exit_log_result(self.signal_sent)
```

When I read documentation for `apt_pkg.get_lock()` via `pydoc3 apt_pkg.get_lock` I see:

> ```
> apt_pkg.get_lock = get_lock(...)
>     get_lock(file: str, errors: bool) -> int
>     
>     Create an empty file of the given name and lock it. If the locking
>     succeeds, return the file descriptor of the lock file. Afterwards,
>     locking the file from another process will fail and thus cause
>     get_lock() to return -1 or raise an Error (if 'errors' is True).
>     
>     From Python 2.6 on, it is recommended to use the context manager
>     provided by apt_pkg.FileLock instead using the with-statement.
> ```

So I think, per above diff, line 342 with condition `if res > 0` should be changed to `if res < 0`.

After modifying that line, I execute:

```
/usr/share/unattended-upgrades/unattended-upgrade-shutdown --wait-for-signal --debug
```

[...]

PR #296, opened 2021-05-03
unattended-upgrade-shutdown: fix wait-for-signal

kucharskim:
I've noticed that unattended upgrade on shutdown doesn't work.
After analyzing the code and some test runs, following cought
my attention, from `pydoc3 apt_pkg.get_lock`:

> apt_pkg.get_lock = get_lock(...)
>     get_lock(file: str, errors: bool) -> int
>
>     Create an empty file of the given name and lock it. If the locking
>     succeeds, return the file descriptor of the lock file. Afterwards,
>     locking the file from another process will fail and thus cause
>     get_lock() to return -1 or raise an Error (if 'errors' is True).
>
>     From Python 2.6 on, it is recommended to use the context manager
>     provided by apt_pkg.FileLock instead using the with-statement.

Based on above, I think condition on line 342 should be reversed.
Running `unattended-upgrade-shutdown --wait-for-signal` with my
diff fixes the issue for me.

This addresses mvo5/unattended-upgrades#295

### DIFF
--- a/unattended-upgrade-shutdown
+++ b/unattended-upgrade-shutdown
@@ -339,7 +339,7 @@ class UnattendedUpgradesShutdown():
         res = apt_pkg.get_lock(self.options.lock_file)
         logging.debug("get_lock returned %i" % res)
         # exit here if there is no lock
-        if res > 0:
+        if res < 0:
             logging.debug("lock not taken")
             if self.lock_was_taken:
                 exit_log_result(self.signal_sent)


kucharskim:
Still something is not exactly right. This needs more investigation.

kucharskim:
I think this is ready for review. I did two runs, one after another, and wanted until the very end. This code makes it work, but someone who knows and understand the code base better should carefully review.

piyaphart90 (approved):

piyaphart90:
- [ ]

rbalint:
I have no idea who `piyaphart90` is and I've reported the user as potentially being a bot.

PR #301, opened 2021-09-01
Add missing options to config files

HendrikVE:
Some options were described in the section "Supported Options Reference" in the README, but were not described in the config files for the different distributions leading to users miss out on these options when they take the default config file as a guide for available options.

Before I had a look at this repository I didn't even know these options existed, because I assumed the config file is "feature-complete". It wasn't, but I think it should be.
It came to my attention, because my mail provider does not consider "root" as a valid "From" field and therefore I never received a mail on upgrades. But of course I did not know that "root" was the default because the option `Unattended-Upgrade::Sender` was missing in my config file (ubuntu).

Btw the option `Unattended-Upgrade::MailReport` was missing on my Ubuntu machine, instead I had the legacy option `Unattended-Upgrade::MailOnlyOnError`, even though `MailReport` is included in the config files in this repository and also handled by the version of `/usr/bin/unattended-upgrade` on my machine (Ubuntu 20.04.3). Is that just `apt` being a bit behind?

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ Supported Options Reference
  packages with the `Unattended-Upgrade::Package-Whitelist-Strict`
  boolean option.
  `Unattended-Upgrade::Package-Blacklist` still applies, thus blacklisted
- packages covered by the whitelist will still not be upraded nor will be
+ packages covered by the whitelist will still not be upgraded nor will be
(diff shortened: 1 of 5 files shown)


chovanecadam:
I have just stumbled across this exact issue. I wish the example configuration file included all default options, or at least a link to this repository which contains all the configuration options.

PR #306, opened 2021-09-30
github: run static checks as part of the workflow

mvo5:
In addition to the full autopkgtest run, also run the "make -C test"
to get quick results.


.github/workflows/build.yml
```diff
@@ -3,10 +3,15 @@ name: CI
...
-  autopkgtest:
+  tests:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+    - name: static checks
+      run: |
+        sudo apt update
+        sudo apt build-dep -y ./
+        make -C test
```
mvo5, line 14: This should now run `pytest-3`

PR #314, opened 2022-02-26
unattended-upgrade: add CM for toggling process priority

jgomo3:
Add a context manager for wrapping code that would benefit
with a different process execution priority, only during that
block of code execution.

It is done in response of a suggestion in the code itself to wrap
some logic in such a context manager.

Then that logic is wrapped in it.

Notice:

Hello @mvo5 . I'm not expecting this PR to be actually accepted at first, I know I haven't wrote tests and I haven't tested it thoroughly. But I found this a good way to communicate with you the idea. Also, I would like to contribute the best I can, just that I would like some pointers on what process do you want me to follow.

I expect that I could iterate over this PR and eventually have a conclusion on this issue.

Thank you for your awesome work.


unattended-upgrade
```diff
@@ -2146,31 +2164,16 @@ def run(options,             # type: Options
...
-    pkgs_to_upgrade = calculate_upgradable_pkgs(cache, options)
-    pkgs_to_upgrade.sort(key=lambda p: p.name)
-    pkgs = [pkg.name for pkg in pkgs_to_upgrade]
-    logging.debug("pkgs that look like they should be upgraded: %s"
-                  % "\n".join(pkgs))
-
-    # FIXME: make this into a ContextManager
-    # stop being nice
-    os.nice(old_priority - os.nice(0))
+    with TryingToBeNince(20):
```
mvo5, line 2167: (nitpick) I wonder if this should just be: `with nice(20):` - I don't have a super strong opinion, I like the fact that it's slightly more verbose and indicates that this might fail. But at the same time short is also nice. Either way, a super small nitpick.

mvo5 (approved):
Thanks, this looks great. Thanks for helping with this and addressing this FIXME in the code.

unattended-upgrade
jgomo3: Cool. `nice` is ok. It is truth that we will loose the indication that it may fail. Maybe `try_nice` could be another option.

So you make the call, any option is good for me.

Aside: how do you test these kind of changes? Do you try in a virtual machine or what? thanks.

mvo5: Hey, sorry that it took me forever to get back to you (life and all that). 

Testing is done via github actions/autopkgtest. It was in a bit of bad shape when you created your original PR. I merged origin/master here now and that should give us working unit tests and end-to-end tests. In geneal having unit tests for this new class would be great.

mvo5: As soon as https://github.com/mvo5/unattended-upgrades/pull/313 is merged testing should be as easy as:
```
$ sudo apt build-dep unattended-upgrades
$ pytest-3
```
(the first step is of course only needed once and a bit of a wart that I need to think about.

mvo5: Fwiw, the code follows more a "foo_bar" style, so "try_nice" (or "nice_try" but only because it's funny) is probably best - also would be nice to debug log any exceptions (actually probably info() them because as root nice normal should work).

mvo5 (commented):
Hey, my apologies for getting back to this so later. I added some inline comments.

unattended-upgrade
jgomo3: I like the "try_nice". Given that the context manager is implemented as a class, the text case conflicts between CamelCase for classess and the snake_case we want for this case.

So, I'm wondering if there is an accepted Pythonic way to write a context manager that can be used in snake_case in the `with` expression.

I know I could hack some function factory, or define a global variable in snake_case for the class. But maybe there is an already accepted convention for Context Managers: are you aware of one such convention?

jgomo3: So the `contextlib` library had the answer: [The contextmanager decorator](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager).

I'm writing it now...

jgomo3: > As soon as #313 is merged testing should be as easy as:
> 
> ```
> $ sudo apt build-dep unattended-upgrades
> $ pytest-3
> ```
> 
> (the first step is of course only needed once and a bit of a wart that I need to think about.

I didn't have `flake8` installed. It is needed for the tests to pass.

mvo5: Thank you for linking to the contextmanager dector docs! Also thanks for updating the code. It looks great, I will merge it as soon as the workflow has finished. I will also add falke8 to the build-dependencies to make sure this works in the future (iirc I switched from pep8/pyflakes to flake8 in one of the recent PRs)

Issue #287, opened 2021-01-19
Support unattended-upgrades specific APT configuration


rbalint:
There are some cases, such as #286 where it would be desired if unattended-upgrades would use a different APT configuration, than APT's command line interface.
The proposed solution is reading config from `/etc/apt/unattended-upgrades.conf` and `/etc/apt/unattended-upgrades.conf.d/` overriding values in `/etc/apt/apt.conf{|.d}`.

julian-klode:
FWIW, APT uses `Binary::<name>::<foo>` options that it then moves into `<foo>` to set different options for different binaries, which might be applicable here too.

rbalint:
@julian-klode Interesting! In case of u-u would the `<name>` be `unattended-upgrade` or `python3` ?

rbalint:
@julian-klode Would you prefer u-u using and documenting the `Binary::<name>` interface or have the separate config (dir)?

duven87:
Hi,
I have been looking for options on how to let update specific packages and I can't find anything.
In my case, I want my servers to only update updates with the security label with the exception of cloud-init which is with label updates.
I have tried: 

"origin=cloudinit";

but it keeps ignoring it.
Is there any idea in the future to add package update exceptions ignoring allow pattern/origin policies?

PR #334, opened 2022-08-29
WIP: Introduce new config options for blacklisting by Origin

### LINKED ISSUE #319: Should not upgrade when pinned to non-allowed origin?
I ran into an issue where u-a upgraded my Firefox in a way I had not expected. The situation is a bit specific, so I'm not sure what the correct behavior would be, but what happens now is at least surprising. Here's the situation.

1. I'm running Ubuntu 22.04 with unattended-upgrades 2.8ubuntu1 and default settings.
2. I added this PPA: https://launchpad.net/~mozillateam/+archive/ubuntu/ppa/+packages
3. I installed firefox from that PPA
4. I set up apt pinning to always prefer firefox from that PPA:

       Package: firefox*
       Pin: release o=LP-PPA-mozillateam
       Pin-Priority: 1000
5. The PPA published a new version, making my locally installed version no longer match the pin (you can reproduce this by installing an older version from the PPA, e.g. https://launchpad.net/~mozillateam/+archive/ubuntu/ppa/+build/23545952`).
6. Unattended-upgrades decides to upgrade the firefox package to the original Ubuntu version, replacing the PPA version.

I thought I had sufficiently configured the system to prefer the PPA version using the pin, but u-a seems to do something different. Given the PPA is not in `Allowed-Origins`, I would not expect the package to be upgraded to the new PPA version, but instead would expect u-a to just leave the package alone in this case. It seems that in practice u-a completely ignores packages outside of its `Allowed-Origins` (selecting the highest version, or highest priority probably, from the set of packages from allowed origins plus the already installed versions, I suppose?).

At step 5 above, here's what I get:

```
matthijs@dottie:~$ apt-cache policy firefox
firefox:
  Installed: 99.0.1+build1-0ubuntu0.22.04.1~mt1
  Candidate: 100.0+build1-0ubuntu0.22.04.1~mt1
  Version table:
     1:1snap1-0ubuntu2 500
        500 http://nl.archive.ubuntu.com/ubuntu jammy/main amd64 Packages
     100.0+build1-0ubuntu0.22.04.1~mt1 1000
        500 https://ppa.launchpadcontent.net/mozillateam/ppa/ubuntu jammy/main amd64 Packages
 *** 99.0.1+build1-0ubuntu0.22.04.1~mt1 100
        100 /var/lib/dpkg/status
     99.0-1 50
         50 http://ftp.nl.debian.org/debian sid/main amd64 Packages
```

From this, I would expect u-a to see that the highest prio version is from the PPA, so outside of its Allowed-Origins, so ignore it.

Instead, u-a chooses to upgrade (see u-a debug output at the end of this post).

```
matthijs@dottie:~$ sudo unattended-upgrade --debug --dry-run &> u-a.txt
```
[u-a.txt](https://github.com/mvo5/unattended-upgrades/files/8592003/u-a.txt)

From the output, here's a bit relevant to the decision to update. It seems that u-a also sees that something is weird (`sanity check failed for: set() : no package is selected to be upgraded or installed`), but maybe I'm misinterpreting that entry.

```
Checking: firefox ([<Origin component:'main' archive:'jammy' origin:'LP-PPA-mozillateam' label:'Firefox ESR and Thunderbird stable builds' site:'ppa.launchpadcontent.net' isTrusted:True>])
sanity check failed for: {'firefox=100.0+build1-0ubuntu0.22.04.1~mt1'} : pkg firefox is not in an allowed origin
falling back to adjusting firefox's dependencies
[...unrelated output about libusb snipped by Matthijs...]
sanity check failed for: set() : no package is selected to be upgraded or installed
pkgs that look like they should be upgraded: firefox
```

I can of course fix this by changing u-a config (add the PPA to `Allowed-Origins` or blacklist the package), but I'd expect this to be not needed (also, I would prefer not spreading around this configuration between u-a config and pinning preferences).

As a workaround, I now added a pin for the non-PPA firefox versions with prio 50 (so it would be less than the 100 for the already installed version), which seems to make u-a not upgrade the package anymore.

```
Package: firefox*
Pin: release o=LP-PPA-mozillateam
Pin-Priority: 1000

Package: firefox*
Pin: release o=Ubuntu
Pin-Priority: 50
```

markkrj:
Fix #319

This allows users to blacklist origins (by pattern) instead of individual packages.
This way, we would not need to change configurations for every package installed from a third party repo, and instead blacklist an entire repo.

I renamed and refactored some functions to better reflect what they do and avoid duplicating code and also because some functions had somewhat misleading/confusing names (as `is_in_allowed_origin`, which checked if a version is in allowed origins and `ver_in_allowed_origin`, which returns a version from a specified set of origin patterns).
If renaming those are not acceptable, or might be breaking, I can revert the naming changes, as it might be merely cosmetic.

I did not added tests and docs yet, as I don't know if this approach would be acceptable.


rbalint:
IMO this should be done by pinning instead and pinning already works reliably.

markkrj:
Imagine we have a third party repo which we use for installing some (not just one) package, and manage updates from it manually. Currently, we would need to:

1. Pin up *every* package coming from 3rd party repo *and* add the repository to allowed origins
   * This would cause u-u to update packages from 3rd party, which might not be desirable
2. Pin down *every* package coming from 3rd party repo to <100 in official repos
   * This require extra work for every package we which to install from 3rd party.

From those options, only 2 would be acceptable, but then we would need to pin every package we install from 3rd parties, and hope to not forget anyone. With the option introduced by this PR, we could pin down an entire repo, from which we intend to manage updates manually.

rbalint:
#287 has a more generic proposed solution for this problem

markkrj:
> #287 has a more generic proposed solution for this problem

How would we solve this problem with APT config?

rbalint:
> > #287 has a more generic proposed solution for this problem
> 
> How would we solve this problem with APT config?

Unattended-upgrades could use additional APT preferences, too, where users could set u-u-specific pinning.

markkrj:
> > > #287 has a more generic proposed solution for this problem
> > 
> > 
> > How would we solve this problem with APT config?
> 
> Unattended-upgrades could use additional APT preferences, too, where users could set u-u-specific pinning.

If the implementation of that proposal respects pinning for disallowed repositories, then yes...

PR #356, opened 2024-02-19
python2 cleaup

a-detiste:
cleanup of remaing Python2 compatibility code.

unittest.mock is already used is some other places.

https://github.com/testing-cabal/mock

> mock is now part of the Python standard library, available as [unittest.mock](https://docs.python.org/dev/library/unittest.mock.html) in Python 3.3 onwards.


mvo5 (approved):
Thank you! This is great!

PR #359, opened 2024-03-31
Correct examples for Debian backports

waterkip:
The examples list a=${distro_codename}-backports as valid, while this should be n=${distro_codename}-backports or
codename=${distro_codename}-backports or a=stable-backports

Closes: #940151


mvo5 (approved):
Thank you!

PR #365, opened 2024-06-22
Install systemd-sleep part into /usr/lib

zeha:
Closes: #1073745

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1073745


mvo5 (approved):
Thank you! Really appreciated!

PR #382, opened 2025-08-26
Document prefix example for Package-Blacklist

Krinkle:
* Explain meaning of existing example.
* Add example that uses a simple prefix.

I'm using this on Debian Trixie. As someone new to this, it took me a while to figure out whether something like `*` (glob/fnmatch style) or `.*` was required. The Debian docs point to this repo.

Ultimately, through the [Ubuntu guide](https://documentation.ubuntu.com/server/how-to/software/automatic-updates/), I figured it should work fine without that. I suggest adding this here to confirm that this is indeed blessed/supported usage, and so that others may find it more easily in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,14 @@ Supported Options Reference
  that it's a list of regular expressions, so you may need to escape special
  characters like "+" as "\\+".
  
- Example:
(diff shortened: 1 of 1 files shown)


mvo5 (approved):
Thank you!

Issue #339, opened 2023-01-26
Feature Request: update old packages only


axet:
Hello!

During recent years I faced multiply failure of system because of automatic update. Some packages which got automatically updated had issues which prevent system to boot or upgrade (in kernel or systemd packages). Having option which allows update only old packages (updates pkg age) would solve those issues.

Can unattended-upgrades check how old update is and only update packages if they are older lets say than 30 days and ignore those updates which just got available?

a-detiste:
You could mess upstream with aptly and cron job and let the leaf nodes untouched.

SveDec:
Hi,

A configuration option to delay upgrades and only upgrade packages older than n days could indeed be very nice, to prevent potential undetected problems from being installed (for instance, I have in mind the 6.1.64-1 kernel problem that happened in December 2023).

Thanks

demccormack:
I would like this too, to protect against supply chain attacks

PR #386, opened 2026-02-12
MinimumAge option

lavr:
This addresses the use case from #339 and helps improve production stability by adding a configurable rollout delay (for example, 7 days) before automatic installation.


mvo5:
Thanks, this is interesting - so do I read this correctly that this reads the {,In}Release file date and only acts if that is older than N (7) days? If so, any "apt update" in between would reset the counter, right? So for this work one needs to make sure to also configure the automatic apt update interval in the apt config and ensure that no manual apt updates happen, right? If I understood this correclty, then I wonder if maybe instead it should work so that it downloads the updates but does not actually install them until they are N (7) days old? This would mean we don't have to worry about spurious apt udpates and the mtime of the deb is our "clock". We still need to worry about the apt autoclean of /var/cache/apt/archives I guess, I have not looked what time we set there. Wdyt?

lavr:
Yes, that’s correct. And indeed, apt update will shift the age. I didn’t account for that, I’ll think about it

PR #387, opened 2026-02-15
MinimumAge option with state tracking and fallback policy

lavr:
This addresses the use case from #339 and helps improve production stability by adding a configurable rollout delay (for example, 7 days) before automatic installation.

Implements persistent seen-version state

On each run, unattended-upgrades updates this state, records new candidate versions with the current date, and prunes stale entries. MinimumAge is then evaluated against this stored first-seen date.


lavr:
I do not like adding state, but on the other hand there is already another state file in the project, so you may consider this type of solution acceptable.

PR #390, opened 2026-04-06
Fix flipped boolean in Config

NWuensche:


mvo5 (commented):
Hey, thanks for your pull-request. The commit message/PR description is a bit sparse, it does not give a rational. I kinda like the current way its done, i.e. the user has to uncomment the option *and* opt-in because its a potentially very disruptive operation. But maybe there is something here I'm missing?

NWuensche:
Hey,

Sorry for the late response. 

I was confused by this option because every
other option around it only has to be uncommented and *not* be opt-in.
But if this is the intention, it's also fine :)

I was just using this option and was debugging some time for why the
option doesn't do anything when you only uncomment it.


Am Do, 09. Apr 05:57, schrieb GitHub Notifications:

PR #392, opened 2026-05-18
Add a new setting Automatic-Reboot-WithSystemUsers

behrmann:
I've run into the problem, that when `Automatic-Reboot-WithUsers` is set to false, so as to not clobber a users running session, reboots will fail, because e.g. `Debian-gdm` is logged in. This is one possible fix. 

Another possible fix would be an explicit list of users to exclude from reboot prevention, which would need more manual work on the distro packaging side, but would be more fine-grained. Happy to rework it such.


mvo5 (approved):
Thank you! I like this, seems very useful. As discussed OOB I would also not mind having an explicit allow-list for users that are allowed on the system when auto-rebooting. But that is orthogonal fine in a separate PR.

behrmann:
Noticed I had confused myself with the logic yesterday and didn't notice with dry-run. I changed this now and factored out the scheduling of the reboot for this. I've manually checked this, but only now noticed the tests. Because they only run for now with approval, I'll need to check this in a Debian environment tomorrow.  I'll also add the missing cases for the tests then, so the whole matrix of user/sysuser logged in or not is covered.

behrmann:
This has now added tests, fixed a few small things I ran into while running the tests, and also adds the option to exclude users from this check, since after the refactor this was a very small addition. PTAL.

data/50unattended-upgrades.Debian
```diff
@@ -119,6 +119,17 @@ Unattended-Upgrade::Package-Blacklist {
...
+// Automatically reboot even if there are system users currently logged in
+// when Unattended-Upgrade::Automatic-Reboot is set to true, but
+// do not reboot if the root user or users with UID larger than 1000 are
+// logged in. This option is ignored if Unattended-Upgrade::Automatic-Reboot-WithUsers
+// is "true"
+//Unattended-Upgrade::Automatic-Reboot-WithSystemUsers "true";
+
+// A comma-separated list of users that are ignored when checking logged in
+// users for Automatic-Reboot-WithUsers and Automatic-Reboot-WithSystemUsers
+//Unattended-Upgrade::Automatic-Reboot-NonBlockingUsers "Debian-gdm,sddm,lightdm";
```
mvo5: Hm, I am a bit unsure about the naming, maybe `Automatic-Reboot-IgnoredUsers` ? But not a super strong opinion, NonBlockingUser is also fine.

mvo5 (commented):
Looks super nice, there is one CI failure unfortuantely it seems

mvo5:
Fwiw, I also relaxed the rules now so that the PR checks should automatically without my approal, that hopefully makes it a bit easier. Sorry for the trouble here.

data/50unattended-upgrades.Debian
behrmann: I thought about `IgnoredUsers` before, but I think it's a bit unclear, but seeing that I didn't manage to write an explanatory text without saying "ignored users" I went with that anyway. Nob-blocking users isn't much clearer either.

behrmann:
> Fwiw, I also relaxed the rules now so that the PR checks should automatically without my approal, that hopefully makes it a bit easier. Sorry for the trouble here.

No worries, my bad for not running the flake test again. At least it noticed I had reused the same name of a test.

behrmann:
@mvo5 Is the failing autopkgtest because of these changes? The warning about the dh-level being too low seems unrelated.

mvo5:
> @mvo5 Is the failing autopkgtest because of these changes? The warning about the dh-level being too low seems unrelated.

I pushed a small tweak, hope this fixes the autopkgtest. The issue is that the default config md5sum is recorded so it needs to get updated. Once green I will merge.

behrmann:
Thanks!

PR #393, opened 2026-05-26
Fix transitive dependency accumulator reuse

Komzpa:
## Summary

- avoid reusing one mutable dependency accumulator across independent transitive dependency walks
- add a focused regression test that calls `transitive_dependencies()` twice with different package dependency sets

## Tests

- `PYTHONPATH=. python3 -m unittest test.test_dependencies test.test_rewind test.test_unavailable_candidate`


mvo5 (approved):
Thanks! Nice catch.

PR #401, opened 2026-05-30
debian: use os-release, drop lsb-release dependency

mvo5:
[draft as it needs #400 first]

This commit switches to parse /etc/os-release for the lsb-release
information. This means we can drop the lsb-release dependency
and avoid the multile forks when getting the DISTRO_ID and friends.

### DIFF
--- a/unattended-upgrade
+++ b/unattended-upgrade
@@ -51,6 +51,7 @@ import re
 import os
 import pwd
 import select
+import shlex
 import signal
 import socket
 import string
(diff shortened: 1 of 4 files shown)


unattended-upgrade
```diff
@@ -113,20 +114,60 @@ SENDMAIL_BINARY = "/usr/sbin/sendmail"
...
-def init_distro_info():
-    global DISTRO_CODENAME, DISTRO_DESC, DISTRO_ID
-    DISTRO_CODENAME = subprocess.check_output(
-        ["lsb_release", "-c", "-s"], universal_newlines=True).strip()
-    DISTRO_DESC = subprocess.check_output(
-        ["lsb_release", "-d", "-s"], universal_newlines=True).strip()
-    DISTRO_ID = subprocess.check_output(
-        ["lsb_release", "-i", "-s"], universal_newlines=True).strip()
+
+def parse_os_release(paths=None):
```
behrmann, line 125: I'm not sure how far back you're targetting `unattended-upgrades`, but since 3.10, which is the oldest supported release by upstream, there is `platform.freedesktop_os_release`, which returns the os-release file as a dict. 

That being said Bullseye is still on 3.9 and Bionic is still on 3.6, so I get if that's not an option.

lewart3: How about the following?

```
123,128c123,137
<     DISTRO_CODENAME = subprocess.check_output(
<         ["lsb_release", "-c", "-s"], universal_newlines=True).strip()
<     DISTRO_DESC = subprocess.check_output(
<         ["lsb_release", "-d", "-s"], universal_newlines=True).strip()
<     DISTRO_ID = subprocess.check_output(
<         ["lsb_release", "-i", "-s"], universal_newlines=True).strip()
---
>     try:
>         # FIXME: move this import to the top of the file
>         import platform
>         info = platform.freedesktop_os_release()
>         DISTRO_CODENAME = info.get("VERSION_CODENAME", "")
>         DISTRO_DESC = info.get("PRETTY_NAME", "Linux")
>         distro_id = info.get("ID", "linux")
>         DISTRO_ID = distro_id[0].upper() + distro_id[1:]
>     except AttributeError:
>         DISTRO_CODENAME = subprocess.check_output(
>             ["lsb_release", "-c", "-s"], universal_newlines=True).strip()
>         DISTRO_DESC = subprocess.check_output(
>             ["lsb_release", "-d", "-s"], universal_newlines=True).strip()
>         DISTRO_ID = subprocess.check_output(
>             ["lsb_release", "-i", "-s"], universal_newlines=True).strip()
```

unattended-upgrade
```diff
@@ -113,23 +114,62 @@ SENDMAIL_BINARY = "/usr/sbin/sendmail"
...
+                if not line or line.startswith("#") or "=" not in line:
+                    continue
+                key, value = line.split("=", 1)
+                # values may be shell-quoted, see os-release(5)
+                try:
+                    tokens = shlex.split(value)
+                except ValueError:
+                    tokens = [value.strip("\"'")]
+                data[key.strip()] = tokens[0] if tokens else ""
+        break
```
lewart3, line 153:
```suggestion
    try:
        # FIXME: move this import to the top of the file
        import platform
        data = platform.freedesktop_os_release()
    except AttributeError:
        for path in (paths or OS_RELEASE_PATHS):
            try:
                f = open(path)
            except FileNotFoundError:
                continue
            with f:
                for line in f:
                    line = line.strip()
                    if not line or line.startswith("#") or "=" not in line:
                        continue
                    key, value = line.split("=", 1)
                    # values may be shell-quoted, see os-release(5)
                    try:
                        tokens = shlex.split(value)
                    except ValueError:
                        tokens = [value.strip("\"'")]
                    data[key.strip()] = tokens[0] if tokens else ""
            break
```

PR #403, opened 2026-06-15
u-u: avoid hangs on big upgrade calculations

### LINKED ISSUE #402: unattended-upgrade infinite loop/hang at 100%CPU
Hi there,

I've recently upgraded to Ubuntu 26.04 which includes unattended-upgrades 2.12ubuntu9 by default (I did not configure it myself nor change anythi0ng in the configuration). This upgrade broke my system insofar as that it 

- regularly hangs at 100% CPU, completely blocking apt in the process
- inhibits system shutdown and suspend-to-RAM

I've had to resort to disabling it completely.

I've attached the output of `unattended-upgrade --dry-run --debug` here and let it run for a bit until I ctrl-Ced. When that runs, it takes 100% CPU. There are a number of held back packages at the moment in my system, but the system is completely updated.

[unattended-upgrade-log.txt](https://github.com/user-attachments/files/28944266/unattended-upgrade-log.txt)

If there is more info I can provide, please let me know.
Thank you,
Johannes

mvo5:
This PR adds some changes to make the upgrade calculation faster and more responsive. 

Closes: https://github.com/mvo5/unattended-upgrades/issues/402

### DIFF
--- a/unattended-upgrade
+++ b/unattended-upgrade
@@ -358,6 +358,11 @@ class UnattendedUpgradesCache(apt.Cache):
     def __init__(self, rootdir):
         # type: (str) -> None
         self._cached_candidate_pkgnames = set()  # type: Set[str]
+        # remember the version adjust_candidate() resolved for a package so we
+        # can re-apply it cheaply after a cache.clear() instead of recomputing
+        # ver_in_allowed_origin(); the result is deterministic for a run
+        self._resolved_candidate_versions = \
+            {}  # type: Dict[str, apt.package.Version]
 
         self.allowed_origins = get_allowed_origins()
         logging.info(_("Allowed origins are: %s"),
@@ -557,6 +562,7 @@ class UnattendedUpgradesCache(apt.Cache):
                     "adjusting candidate to version %s in allowed "
                     "origin %s" % (new_cand, new_cand.origins))
                 pkg.candidate = new_cand
+                self._resolved_candidate_versions[pkg.name] = new_cand
                 return True
             else:
                 return False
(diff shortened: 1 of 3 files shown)


unattended-upgrade
```diff
@@ -1569,6 +1590,28 @@ def rewind_cache(cache, pkgs_to_upgrade):
...
+            pkg2.mark_install(from_user=not pkg2.is_auto_installed)
+        fast_path_ok = (
+            cache.broken_count == 0
+            and sanity_problem(cache, None) is None
+            and all(pkg2.marked_install or pkg2.marked_upgrade
+                    for pkg2 in pkgs_to_upgrade))
+    except SystemError:
+        fast_path_ok = False
+    if fast_path_ok:
+        return
```
behrmann: I think you could turn down the syntax here a bit by dropping the `fast_path_ok` variable, since it's not used in the previous path
```suggestion
        if (
            cache.broken_count == 0
            and sanity_problem(cache, None) is None
            and all(pkg2.marked_install or pkg2.marked_upgrade
                    for pkg2 in pkgs_to_upgrade)):
            return
    except SystemError:
        pass
```

unattended-upgrade
```diff
@@ -1317,9 +1335,12 @@ def check_changes_for_sanity(cache, desired_pkg=None):
     if sanity_check_result is None:
         return True
     else:
-        logging.debug("sanity check failed for: %s : %s"
-                      % (str({str(p.candidate) for p in cache.get_changes()}),
-                         sanity_check_result))
+        # the change set behind a failed check can be hundreds of
+        # packages: so only the number and the failed pkg
```
behrmann:
```suggestion
        # packages: so only show the number and the failed pkg
```

PR #406, opened 2026-06-26
feat: include esm legacy {infra,apps} in Ubuntu conf

srunde3:
When available, include packages from the legacy infra and apps repositories for Ubuntu distributions. This ensures continuity of support once regular ESM coverage reaches end-of-life.


mvo5 (approved):
Thanks

